PR #2, opened 2022-06-20
feat: add nix flake package manager support

thenonameguy:
I'm very impressed with the work so far on this repo :tada: ... so I decided to add Nix Flake support.
I think `process-compose` could be a missing link in the Nix ecosystem, since with the combination with projects like https://numtide.github.io/devshell/
the necessary binaries for the services can be provided consistently for awesome OCI-less developer environments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ coverage.out
 bin/**
 *.log
 .env
+# local build output via go build
+process-compose
+
+# nix build .#process-compose output
+result
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,21 @@
+{ config, lib, pkgs, ... }:
+
+pkgs.buildGoModule rec {
+  pname = "process-compose";
+  version = "0.10.2";
+  src = ./.;
+  ldflags = [ "-X main.version=${version}" ];
+
+  vendorSha256 = "1syn7sfv2hqwyl16kg14rvgwil65zlabnv2g44bpi4q35xmv1q46";
+
+  postInstall = "mv $out/bin/{src,process-compose}";
+
+  meta = with lib; {
+    description =
+      "Process Compose is like docker-compose, but for orchestrating a suite of processes, not containers.";
+    homepage = "https://github.com/F1bonacc1/process-compose";
+    license = licenses.asl20;
+    mainProgram = "process-compose";
+  };
+  doCheck = false; # it takes ages to run the tests
+}
--- a/README.md
+++ b/README.md
@@ -8,17 +8,15 @@ Process Compose is a simple and flexible scheduler and orchestrator to manage no
 
 Main use cases would be:
 
-* Processes execution (in parallel or serially)
-* Defining processes dependencies and startup order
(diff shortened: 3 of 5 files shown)


default.nix
```diff
@@ -0,0 +1,21 @@
+{ config, lib, pkgs, ... }:
+
+pkgs.buildGoModule rec {
+  pname = "process-compose";
+  version = "0.10.2";
```
F1bonacc1, line 5: Do you think there is a way to take the version value from the latest git tag?
If not I think I can add some sort of automation to the Makefile or create a git hook.

.gitignore
```diff
@@ -3,3 +3,8 @@ coverage.out
 bin/**
 *.log
 .env
+# local build output via go build
+process-compose
```
F1bonacc1, line 7: Doesn't the output go to the bin directory?

F1bonacc1 (commented):
Thank you for the PR. Everything looks great.
Just 2 small comments and we are good to go.

default.nix
thenonameguy: It's usually hand-managed in cases where the source of the repository/package is remote. Since it's local, you could create a simple version.txt file with some script (outside of Nix) and reference it here with 'version = ./version.txt'

.gitignore
thenonameguy: No, it gets symlinked to ./result. I just tried out some of the build commands manually and figured it's safer to prevent checking in the binary accidentally. Absolutely can be removed.

default.nix
F1bonacc1: Will this change work with both remote and local packages? Sorry if it sounds like a trivial question, I am just not familiar with the nix syntax.
Will the versions file have to be committed as well? Is there any special syntax for the contents or any "version like" string will work (e.g. v0.10.3)?

thenonameguy: The way it's written now it uses the local filesystems contents to build the project.
In the end if some1 wants to build it by fetching the repo via git, they can do it like this (found a random package a definition on the internet):
```nix
pet = buildGoModule rec {
  pname = "pet";
  version = "0.3.4";

  src = fetchFromGitHub {
    owner = "knqyf263";
    repo = "pet";
    rev = "v${version}";
    sha256 = "0m2fzpqxk7hrbxsgqplkg7h2p7gv6s1miymv3gvw0cz039skag0s";
  };

  vendorSha256 = "1879j77k96684wi554rkjxydrj8g3hpp0kvxz03sd8dmwr3lh83j";

  meta = with lib; {
    description = "Simple command-line snippet manager, written in Go";
    homepage = "https://github.com/knqyf263/pet";
    license = licenses.mit;
    maintainers = with maintainers; [ kalbasit ];
  };
}
```
> Will the versions file have to be committed as well?
Yes, it's best practice if you want to go this route.
> Is there any special syntax for the contents or any "version like" string will work (e.g. v0.10.3)?
Generally the version doesn't have a `v` prefix, but see for yourself:
https://github.com/NixOS/nixpkgs/search?q=version+%3D

Issue #3, opened 2022-06-22
Feature request: per-process shutdown command / trap support


thenonameguy:
For my use-case most of the commands I'm managing via `process-compose` eventually yield a process into the background, and exit cleanly with 0, with a no retry policy.
Other processes use the `process_completed` dependency feature to realize that the given process (let's say a postgres init script) has concluded and now the postgres server is ready in the background.
So far so good :tada: 

Now, the only problem with these daemon-launching scripts are that if the `process-compose` process is killed, the underlying daemons are not turned off.
This leaves the environment in an inconsistent state (unlike calling `docker-compose down`/ Ctrl-C'ing the `up` command).

Proposal:

- add `processes.$PROCESSNAME.shutdown_command: string` to the schema
- handle SIGTERM in `process-compose` itself, `exec.Command` these commands in parallel, wait for the command to terminate.

stretch goal/kinda unrelated but-nice-to-have:
- revisit the current SIGKILL way of killing the underlying process (instead use SIGTERM), so the command has a chance to run any kind of [bash trap](https://www.linuxjournal.com/content/bash-trap-command) like functionality

![image](https://user-images.githubusercontent.com/2217181/174923929-41906caf-dc84-427d-a7ce-d60ed158dd7b.png)

F1bonacc1:
I think that the proposal is clear:

1. On process termination, if `processes.$PROCESSNAME.shutdown_command` is defined, execute the command
2. Wait for `processes.$PROCESSNAME.shutdown_timeout` amount of seconds for the command to return
3. In case of timeout - sigkill
4. If `processes.$PROCESSNAME.shutdown_timeout` is not defined use 10sec?

Does it sound reasonable?

thenonameguy:
> I think that the proposal is clear:
> 
> 
> 
> 1. On process termination, if `processes.$PROCESSNAME.shutdown_command` is defined, execute the command
> 
> 2. Wait for `processes.$PROCESSNAME.shutdown_timeout` amount of seconds for the command to return
> 
> 3. In case of timeout - sigkill
> 
> 4. If `processes.$PROCESSNAME.shutdown_timeout` is not defined use 10sec?
> 
> 
> 
> Does it sound reasonable?

Perfect, good catch with not waiting indefinitely! 10s seems plenty, but not too much.

F1bonacc1:
[...]

PR #5, opened 2022-06-25
perform graceful termination of process compose

F1bonacc1:


F1bonacc1:
This should finalize the #3 issue

src/app/process.go
```diff
@@ -157,12 +164,20 @@ func (p *Process) WaitForCompletion(waitee string) int {
...
 
-func (p *Process) WontRun() {
+func (p *Process) wontRun() {
 	p.onProcessEnd()
 
 }
 
+// perform gracefull process shutdown if defined in configuration
 func (p *Process) shutDown() error {
+	if !p.isState(ProcessStateRunning) {
```
thenonameguy, line 174: This check is messing my use-case up.
With the `pg_ctl start` example, the process goes to `completed` state, since it exits with 0 status code.
In this case the shutdown command not run.
For the sake of simplicity, I think we can ask users to provide an idempotently runnable / non-failing command like:
```
pg_ctl stop || true
```
and run it regardless of process state.

PR #9, opened 2022-07-20
fix(nix): adjust nix build instructions, update dependency sha

thenonameguy:
NOTE: every time you change the dependencies of the project, the nix build breaks (as the dependencies need to be reproducible).
Consider including nix in the build process (better yet, build the final static Go binaries for the releases using it).

I'll keep doing this until we are using this project though :)

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -6,7 +6,7 @@ pkgs.buildGoModule rec {
   src = ./.;
   ldflags = [ "-X main.version=${version}" ];
 
-  vendorSha256 = "1syn7sfv2hqwyl16kg14rvgwil65zlabnv2g44bpi4q35xmv1q46";
+  vendorSha256 = "RcxU68vn2RirbOtEGcVrKYmEFeQDg1eovlQKXRandNs=";
 
   postInstall = "mv $out/bin/{src,process-compose}";
 
--- a/flake.nix
+++ b/flake.nix
@@ -8,12 +8,21 @@
 
   outputs = { self, nixpkgs, flake-utils }:
     flake-utils.lib.eachDefaultSystem (system:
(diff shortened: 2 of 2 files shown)


thenonameguy:
Sure, the process is quite simple:
1. Install the nix package manager with flake support (the latest version already has everything out-of-the-box) via:
   https://nixos.org/download.html#download-nix 
2. invoke `nix build github:F1bonacc1/process-compose`
   To understand final argument  I recommend reading `nix flake --help`, especially the Examples section
   To repro the failure of not being able to run/build the project prior this PR you need to run the exact SHA as a base:
   ```
   nix build github:F1bonacc1/process-compose/d8f1fb9af7b5ed00c934eac6d08d79de8935af4c
error: hash mismatch in fixed-output derivation '/nix/store/6nwvsfr66iqgi69gg41czdrw0g386ngf-process-compose-0.16.0-go-modules.drv':
         specified: sha256-huCway8Dk3gXIU9suxT9xdDI384kvGkC9RxDsZ0+1us=
            got:    sha256-RcxU68vn2RirbOtEGcVrKYmEFeQDg1eovlQKXRandNs=
error: 1 dependencies of derivation '/nix/store/n7zcvj9c5pw3ns43y3xgmcyaf9ak52qa-process-compose-0.16.0.drv' failed to build
```
Now you can see that the intermediate `go-modules` dependencies should have produced a certain hash, but it didn't (hence the runtime error while trying to build the project). Now my diff in this PR should make sense.

When you are locally developing, using `nix build .` or `nix run .` will do the things you expect, using the local source code/dependency data from the project. This will surface any differing SHA's, before releasing a new version.
The price you pay for better protection from supplier-chain-attacks :)

PR #16, opened 2022-10-23
15 expose cli rpc in addition to http

F1bonacc1:


src/cmd/restart.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright © 2022 NAME HERE <EMAIL ADDRESS>
+*/
```
thenonameguy, line 3: Is this intentional?

thenonameguy (changes requested):
Double check the copyright header in all new files, maybe replace with actual license of the project. Feature looks great!

src/cmd/restart.go
F1bonacc1: Missed it! Thanks for pointing this out :)

Issue #14, opened 2022-10-19
Feature Request: Run a "sanity-check" program


IngwiePhoenix:
Hello!

This is a very great project! I have been looking for a much smaller but still capable process supervisor compared to PM2 and this looks like it is exactly what i need :)

I would like to request/suggest a feature: Allow the user to specify a "sanity check". For instance, imagine this shell script:

```sh
#!/bin/bash

check() {
  command -v $1 >/dev/null
  if [ $? != 0 ]; then
    echo "$1 was not found!"
    exit 1
  fi
}

check go
check node
```

If either of these fail, the script exits with exit code 1.

What I would like to do is to run this before any of the processes are invoked as a way to introduce a script that verifies the environment if everything is exactly where it should be. :)

Again thanks a lot for this project, been playing around with it today a lot and it's great!

Kind regards,
Ingwie

IngwiePhoenix:
Good morning!

You are most welcome!

1. Globally. It should just make sure all things are where and how they should be.
2. No, it should run before all processes. If it returns true, assume that everything is fine and proceed to start the processes.
3. This could be considered - but I think it'd be better if a per-process check failed, it should only mean that this process can not be started and should only affect dependees.
4. Depends on if the user supplied arguments or not. Like if they only want to run one process out of many, I think the global check should still run. You could consider passing a `PROCESS_COMPOSE_PROCS` environment variable to tell the check if it should limit its checks or not.

Feel free to ask! I am not very good with Go, so there is not that much I can contribute. More of a C and Nodejs guy. ^^;

F1bonacc1:
Good afternoon.

I think I have a clearer picture of what you want to achieve.
I think that even today this is almost possible if you define the first process to be this check and all other processes are dependent on it. Or at least the first process in the chain is dependent on it.

The only thing that is missing is the ability to exit on process failure.
What do you think about the following:

1. I will add support to configure an exit strategy if the process exit code != 0. This will be optional. If not configured the behavior will be as today.
2. If configured and exit code != 0 the process-compose will gracefully terminate all the other running processes (in case of the preflight check there shouldn't be any) and exit with the same exit code as the check process.
3. As today, you will be able to disable/enable the check process in configuration.
4. As a separate feature, I will add the support to run specific processes by providing a CLI argument to process compose. This way you will be able to run all the checks individually and validate their correctness before running all the other processes.
[...]

PR #28, opened 2022-11-19
Rework nix instructions now that nixpkgs has it

thenonameguy:
now that this project is packaged in the [main nixpkgs repostitory](https://github.com/NixOS/nixpkgs/pull/201877)
the need for instructing people to compile from source is unnecessary.

I'm also proposing changes to include process-compose as a built-in component in this project:
https://devenv.sh/
https://github.com/cachix/devenv/issues/78

The above would solve problems like https://github.com/F1bonacc1/process-compose/issues/14
and provide nice common service descriptions for common architectural components (postgres/kafka/redis/etc.).

Might be worth mentioning in the README.md once the above issue is resolved.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,8 +1,8 @@
-{ config, lib, pkgs, installShellFiles, ... }:
+{ buildGoModule, config, lib, pkgs, installShellFiles, version ? "latest" }:
 
-pkgs.buildGoModule rec {
+buildGoModule {
   pname = "process-compose";
-  version = "0.24.1";
+  version = version;
   src = ./.;
   ldflags = [ "-X main.version=v${version}" ];
 
--- a/README.md
+++ b/README.md
@@ -111,7 +111,11 @@ process-compose -f /path/to/process-compose.yaml
 - If you have the Nix package manager installed with Flake support, just run:
 
 ```sh
-nix run github:F1bonacc1/process-compose
+# to use the latest binary release
+nix run nixpkgs/master#process-compose -- --help
+
+# or to compile from the latest source
+nix run github:F1bonacc1/process-compose -- --help
 ```
 
 ### Documentation
--- a/flake.nix
+++ b/flake.nix
@@ -15,7 +15,9 @@
         };
       in {
         overlays.default = final: prev: {
-          process-compose = final.callPackage ./default.nix { };
+          process-compose = final.callPackage ./default.nix {
+            version = self.shortRev or "dirty";
+          };
         };
         overlay = self.overlays.default;
         packages = { inherit (pkgs) process-compose; };


default.nix
```diff
@@ -1,8 +1,8 @@
-{ config, lib, pkgs, installShellFiles, ... }:
+{ buildGoModule, config, lib, pkgs, installShellFiles, version ? "latest" }:
```
F1bonacc1, line 1: Quick question about this line.
There is a mechanism in PC which checks (on startup) if the user runs the latest version.
Basically, it compares the current version to the latest release in GitHub.
If PC will be compiled with the string "latest" instead of the real version, it will always show that there is a pending update.
Is there any way around this issue?

thenonameguy: Yup, great question.
So the point is you are now going to want to use the binary compiled by the `nixpkgs` repo from now on.

That will have the version available as defined in https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/misc/process-compose/default.nix

The package will show up here in a few days once the unstable channel "stabilizes":
https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=process-compose 

The only reason you would want to go through this way of building the software is to:
1. build a custom commit from source
2. use an unreleased version of the project (the above nixpkgs file is not recent enough)

In both cases you probably don't care about the newer versions, as this is an 'advanced' use-case.

I can see the purpose of the newer version check, but breaking it for 'custom' builds sounds acceptable to me.
Other software usually doesn't have this as prominent in the UI as `process-compose` has, I'm questioning the need for it as the software matures (`docker-compose` does not warn you about using an outdated version since years).

F1bonacc1: I will start with the last part. The reason I added this feature is that I saw and liked a similar functionality in K9S. Also, it is very common (e.g. all the chromium based browsers, all the JetBrains IDEs, VsCode) to show this functionality in the UI.

If the issue is limited to custom builds, then I think that it is acceptable.
Is there a scenario in which I release a new version and someone who installed PC through nix will see this new version notification, but won't be able to update because there is a delay in updating the nix packages repository?
Will this repository be updated automatically or there is a manual process involved in bringing it to the main repo?

thenonameguy: > I will start with the last part. The reason I added this feature is that I saw and liked a similar functionality in K9S. Also, it is very common (e.g. all the chromium based browsers, all the JetBrains IDEs, VsCode) to show this functionality in the UI.

Sounds reasonable :tada: , I'm personally not using those programs daily, so I haven't been exposed to this pattern.

> Is there a scenario in which I release a new version and someone who installed PC through nix will see this new version notification, but won't be able to update because there is a delay in updating the nix packages repository?

Yes, this is quite common, as the updating frequency of the unstable channel is a few days usually.
This is not a unique problem to Nix though, image if Debian stable packaged PC and you'd do weekly releases.
Users would see constantly the out-dated warning in the TUI. 
VSCode/Chromium have self-updating capabilities to resolve these issues, although they beat the atomic upgrade features of package managers.
Tradeoffs, tradeoffs, tradeoffs..

> Will this repository be updated automatically or there is a manual process involved in bringing it to the main repo?

It's a manual effort. As the package maintainer in the `nixpkgs` repo, I'm responsible to keep updating the version/dependency hashes. This can be done by anyone though, taking k9s in nixpkgs as an example:
https://github.com/NixOS/nixpkgs/commits/master/pkgs/applications/networking/cluster/k9s/default.nix

F1bonacc1: How about this:
Nix users won't care about this notification in any case because of 2 reasons:
1. There is nothing to do about it.
2. PC will be updated automatically when they run the system update command (I believe this one, right? :))

Github downloaders will still get this feature.

The way to distinguish between the 2 is to use different ldflags in the build process.
In my `Makefile` I will set the `version` variable and in `default.nix` I will set the `nixVersion` variable. The `version` will remain `undefined`, which will let me know not to pull the latest release from GitHub.

What do you think?

thenonameguy: I like the idea!

I would just replace `Nix` with `package manager` users in general in your comment.

> In my Makefile I will set the version variable and in default.nix I will set the nixVersion variable. The version will remain undefined, which will let me know not to pull the latest release from GitHub.

No need for setting `nixVersion` specific ldflag. Instead let's add an auto-update check specific boolean flag.
This way people who are source-builders and don't want network requests on TUI startup to happen can opt-out of it.
I'll modify the packaging source code in nixpkgs to set this to false (and also in this PR).

Does that sound okay?

F1bonacc1: Perfect!

F1bonacc1:
Hey @thenonameguy,

I gave it a little more thought. You don't need to change your PR. I will initialize the boolean flag to be false, so if you pass nothing we will get the desired behavior.
I will update my makefile to pass true, of course.

PR #36, opened 2022-12-24
Added version command - prints process-compose version and build info

F1bonacc1:
Hi @thenonameguy 

I was trying to add a version info command that prints the following output:
```
Version:        v0.29.1
Commit:         00a3211
Date (UTC):     2022-12-24T21:19:13Z
```
I did some changes to `default.nix` and `flake.nix`, but it still doesn't work as expected:
```
Version:        v0.29.1
Commit:         dirty
Date (UTC):     20221224212048
```
Can you please give me a hint or send me to the documentation on how to do it properly?
Also, I made some changes to `default.nix` to make it more like: 
https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/misc/process-compose/default.nix

Can you please confirm that it's indeed the right direction?

Thanks!

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -29,7 +29,11 @@ builds:
         goarch: arm
     dir: src
     ldflags:
-      - -X github.com/f1bonacc1/process-compose/src/config.Version={{.Tag}} -X github.com/f1bonacc1/process-compose/src/config.CheckForUpdates=true -s -w
+      - -X github.com/f1bonacc1/process-compose/src/config.Version={{.Tag}}
+      - -X github.com/f1bonacc1/process-compose/src/config.CheckForUpdates=true
+      - -X github.com/f1bonacc1/process-compose/src/config.Commit={{.ShortCommit}}
+      - -X github.com/f1bonacc1/process-compose/src/config.Date={{.Date}}
+      - -s -w
 archives:
   - replacements:
       darwin: Darwin
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -32,7 +32,7 @@ builds:
       - -X github.com/f1bonacc1/process-compose/src/config.Version={{.Tag}}
       - -X github.com/f1bonacc1/process-compose/src/config.CheckForUpdates=true
       - -X github.com/f1bonacc1/process-compose/src/config.Commit={{.ShortCommit}}
-      - -X github.com/f1bonacc1/process-compose/src/config.Date={{.Date}}
+      - -X github.com/f1bonacc1/process-compose/src/config.Date={{.CommitDate}}
       - -s -w
 archives:
   - replacements:
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ NAME=process-compose
 RM=rm
 VERSION = $(shell git describe --abbrev=0)
 GIT_REV    ?= $(shell git rev-parse --short HEAD)
-DATE       ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
+DATE       ?= $(shell TZ=UTC0 git show --quiet --date='format-local:%Y-%m-%dT%H:%M:%SZ' --format="%cd")
 NUMVER = $(shell echo ${VERSION} | cut -d"v" -f 2)
 PKG = github.com/f1bonacc1/${NAME}
 SHELL := /bin/bash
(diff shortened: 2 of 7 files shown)


.goreleaser.yaml
```diff
@@ -29,7 +29,11 @@ builds:
         goarch: arm
     dir: src
     ldflags:
-      - -X github.com/f1bonacc1/process-compose/src/config.Version={{.Tag}} -X github.com/f1bonacc1/process-compose/src/config.CheckForUpdates=true -s -w
+      - -X github.com/f1bonacc1/process-compose/src/config.Version={{.Tag}}
+      - -X github.com/f1bonacc1/process-compose/src/config.CheckForUpdates=true
+      - -X github.com/f1bonacc1/process-compose/src/config.Commit={{.ShortCommit}}
+      - -X github.com/f1bonacc1/process-compose/src/config.Date={{.Date}}
```
thenonameguy, line 35: This is a misnaming, dates are generally referring to days, instead of timestamps when the build happened, which is the actual case.

Makefile
```diff
@@ -1,11 +1,16 @@
 NAME=process-compose
 RM=rm
 VERSION = $(shell git describe --abbrev=0)
+GIT_REV    ?= $(shell git rev-parse --short HEAD)
+DATE       ?= $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
```
thenonameguy, line 5: Again, as above, this is not a date, but a build timestamp.

default.nix
```diff
@@ -1,10 +1,18 @@
-{ buildGoModule, config, lib, pkgs, installShellFiles, version ? "latest" }:
+{ buildGoModule, config, lib, pkgs, installShellFiles, date, commit }:
 
-buildGoModule {
+buildGoModule rec {
   pname = "process-compose";
-  version = version;
+  version = "0.29.1";
+  pkg = "github.com/f1bonacc1/process-compose/src/config";
```
thenonameguy, line 6: This does not work as a the `buildGoModule` has a limited attribute set (= hash-maps/key-value store) that it accepts. A let binding is more fortunate, see my follow-up PR.

default.nix
```diff
@@ -21,9 +29,9 @@ buildGoModule {
   '';
 
   meta = with lib; {
-    description =
-      "Process Compose is like docker-compose, but for orchestrating a suite of processes, not containers.";
+    description = "A simple and flexible scheduler and orchestrator to manage non-containerized applications";
     homepage = "https://github.com/F1bonacc1/process-compose";
+    changelog = "https://github.com/F1bonacc1/process-compose/releases/tag/v${version}";
```
thenonameguy, line 34: :pinched_fingers:  :tada:

thenonameguy (changes requested):
Please pull in https://github.com/thenonameguy/process-compose/tree/version_info to fix most of the problems.
See the comments otherwise.
Including the last modified timestamp makes it a bit more brittle, once you have a commit that's merged, the `commit` will be not `dirty`, but the SHA.
Nix has no built-in timestamp parsing fns to have a consistent value between `go install` and `nix build`. For now IMO it works as it is.

.goreleaser.yaml
F1bonacc1: I can change it to:
`github.com/f1bonacc1/process-compose/src/config.Date={{.CommitDate}}`
Go releaser support that out of the box (nice) and returns `2022-12-24T21:20:48Z`
And in Makefile I change it to:
`DATE       ?= $(shell TZ=UTC0 git show --quiet --date='format-local:%Y-%m-%dT%H:%M:%SZ' --format="%cd")`
Which also returns the last commit timestamp in the same format: `2022-12-24T21:20:48Z`

It's also somewhat supported with flake:
`self.lastModifiedDate`
The format is a little bit different, but also in UTC: `20221224212048`

However no matter what I tried I can't make a short commit work, it always returns dirty. Even after taking your changes.

thenonameguy: If you have a clean git status, it works properly:
```
➜  process-compose git:(version_info_upstream) git status                           
On branch version_info_upstream
nothing to commit, working tree clean
➜  process-compose git:(version_info_upstream) nix build .                         
➜  process-compose git:(version_info_upstream) ./result/bin/process-compose version
Process Compose
Version:        v0.29.1
Commit:         89254f1
Date (UTC):     20221225200050
License:        Apache-2.0

Written by Eugene Berger
```
I haven't found a nice and easy way to parse the `lastModifiedDate` into something a bit more `date` like without too much hassle. IMO the way this PR stands in fine for now.

PR #73, opened 2023-07-19
config: add process.shutdown.only_kill_process option

### LINKED ISSUE #67: Add an option to not send the shutdown signal to a whole process group.
## Feature Request

Add an option to send the shutdown signal only to the process started by `process-compose` instead of its whole process group.

#### Use Case:

Consider the following Python program (Linux 6.1, Python 3.11):

```python
import os
import signal
import threading
import multiprocessing as mp


stop = threading.Event()


def handler(signum, _frame):
    print(f"Process {os.getpid()} received signal {signum}.", flush=True)
    stop.set()


if __name__ == "__main__":
    print("main:", os.getpid())

    signal.signal(signal.SIGTERM, handler)
    
    ctx = mp.get_context("spawn")
    with ctx.Pool(processes=1) as _pool:
        stop.wait()

    print("Pool terminated", flush=True)

```

This is a small variation on the first example of the documentation of the [`multiprocessing`](https://docs.python.org/3/library/multiprocessing.html?highlight=Pool#introduction) module.

The relevant part of the process tree is:

```
  PGID     PID    PPID COMMAND
 177518  177518   98725 /usr/bin/python proc.py
 177518  177542  177518  \_ /usr/bin/python -c from multiprocessing.resource_tracker import main;main(5)
 177518  177543  177518  \_ /usr/bin/python -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=6, pipe_handle=12) --multiprocessing-fork
```

When terminated with

```bash
kill -TERM 177518
```

the process terminates properly as expected:

 ```
main: 177518
Process 177518 received signal 15.
Pool terminated
```

However, when the termination signal is sent to the whole process group:

```bash
kill -TERM -177518
```

the `Pool`'s context hangs on exit. My understanding is that `Pool.__exit__` (the method called when exiting the context) calls [`Pool.terminate`](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.terminate), which hangs because the associated process was already terminated by the signal sent to the process group.

Since, on UNIX, `process-compose` always sends the registered shutdown signal to the whole process group ([`process_unix.go:23`](https://github.com/F1bonacc1/process-compose/blob/303222889c1f211ac005575f0888e76b663d551e/src/app/process_unix.go#L23)), such a program cannot be properly terminated properly by `process-compose`.

#### Proposed Change:

Add an option to the `shutdown` section of the configuration schema to send the shutdown signal to the managed process' PID only, instead of the whole process group.

For example:

```yaml
shutdown:
    signal: 15
    # ...
    kill_process_group: yes  # default to keep backwards compatibility
```

Processes spawned by `process-compose` that actively manage their children processes.
I feel like the current approach is quite invasive since it also circumvents any shutdown ordering the parent process might want to do on the children.

#### Alternative Approaches

This may well be a bug/feature of the Python standard library. However, while killing the whole process group makes sense when mimicking  what `docker-compose` does (this seems to be the spirit of #49), a less aggressive shutdown strategy is more predictable (and would reflect better what the readme says: "In case only shutdown.signal is defined [1..31] the running process will be terminated with its value.").

airwoodix:
`shutdown.only_kill_process` defaults to false (backwards compatible). In this case and on UNIX platforms, the shutdown signal is sent to the running process' process group. If `shutdown.only_kill_process` is true, the shutdown signal is only sent to the running process itself.

Closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -245,17 +245,20 @@ nginx:
     command: "docker stop nginx_test"
     timeout_seconds: 10 # default 10
     signal: 15 # default 15, but only if the 'command' is not defined or empty
+    parent_only: no  # default no. If yes, only signal the running process instead of its whole process group
 ```
 
-`shutdown` is optional and can be omitted. The default behavior in this case: `SIGTERM` is issued to the running process.
+`shutdown` is optional and can be omitted. The default behavior in this case: `SIGTERM` is issued to the process group of the running process.
 
-In case only `shutdown.signal` is defined `[1..31] ` the running process will be terminated with its value.
+In case only `shutdown.signal` is defined `[1..31] ` the running process group will be terminated with its value.
+
+If `shutdown.parent_only` is yes, the signal is only sent to the running process and not to the whole process group.
 
 In case the `shutdown.command` is defined:
 
 1. The `shutdown.command` is executed with all the Environment Variables of the primary process
 2. Wait for `shutdown.timeout_seconds` for its completion (if not defined wait for 10 seconds)
-3. In case of timeout, the process will receive the `SIGKILL` signal
+3. In case of timeout, the process group will receive the `SIGKILL` signal (irrespective of the `shutdown.parent_only` option).
 
 ##### Background (detached) Processes
 
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -248,7 +248,8 @@ func (p *Process) shutDown() error {
 	if isStringDefined(p.procConf.ShutDownParams.ShutDownCommand) {
 		return p.doConfiguredStop(p.procConf.ShutDownParams)
 	}
-	return p.stop(p.procConf.ShutDownParams.Signal)
+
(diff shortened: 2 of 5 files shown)


F1bonacc1:
Thank you for your PR @airwoodix!

It looks good!
A small comment. `only_kill_process` sounds a bit ambiguous to me and it implies `kill` as in `kill -9`.
What do you think about  `shutdown_parent_only`?

airwoodix:
Thanks for the quick and positive review @F1bonacc1!

I renamed the option `shutdown.parent_only` (to avoid the `shutdown` repetition). Does this make sense to you?

The change made the logic in the `stop` method more straightforward. Also enhanced the new log line there and ran `gofmt`.

F1bonacc1:
Your suggestion is even better 🙂
Merging.

PR #74, opened 2023-07-19
Escape the TUI log lines in ANSI color mode

enobayram:
This PR escapes the log line in the `useAnsi` branch of `LogView.WriteString`. Without this escape, substrings that match `[\w+]` are interpreted by regions and cause display issues.


enobayram:
This PR can be tested by running process compose with the following config:
```

F1bonacc1:
Good catch @enobayram!

Thank you for the PR!

enobayram:
And thank you for reviewing and merging it quickly! (Not to mention developing this excellent utility :slightly_smiling_face:)

PR #76, opened 2023-07-20
logger: create parent directories if necessary

### LINKED ISSUE #71: Create intermediate directories of log_location if they do not exist
## Feature Request

I'd like process-compose to create intermediate directories to log_location.
Currently setting log_location to within a missing directory will result in error on startup.  

#### Use Case:

Improving DX when using custom log locations. Currently this may require additional steps or wrapper script when running given configuration for the first time.

#### Proposed Change:

create intermediate directories for all log_location's

people using custom log_location

#### Alternative Approaches

wrapper script or extra manual steps which need to be documented

airwoodix:
Closes #71

### DIFF
--- a/src/pclog/logger_facade.go
+++ b/src/pclog/logger_facade.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"github.com/rs/zerolog/log"
 	"os"
+	"path"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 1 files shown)


adrian-gierakowski:
Awesome, thanks!

F1bonacc1:
Thanking for taking care of this!
It's a new and welcome trend 😃

PR #78, opened 2023-07-21
feat: availability.exit_on_end

### LINKED ISSUE #77: add option to stop all other process on one process completion
## Feature Request

Similar to `exit_on_failure` but also exists if give process exits with 0 status

#### Use Case:

Running integration tests, where multiple supporting services (databases etc.) need to be started first, followed by the test executor process. Once test executor completes (regardless of exit code), all other processes should be shut down and PC should exit with the code returned by test executor.

#### Proposed Change:

add boolean flag: `availability.exit_on_end`
the reason why this is not just another option in `availability.restart` is  that you might want to restart a process on failure up to certain amount of times, and only once retries are exhausted (or once process completes successfully) shut down all processes.

Anyone which wishes to run one-shot a process which need other processes to run in the background while the main process is executing. 

#### Alternative Approaches

One could execute `process-compose process stop PROCESS_NAME` for all supporting processes once the main process finished

adrian-gierakowski:
a bool flag, if set to true, shuts down all processes once
given process ends (regardless of exit code)

this can be used together with `availability.restart = on_failure` to
have the process restarted on failure and only exit PC on success

closes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,7 +474,7 @@ process2:
 
 ##### Terminate Process Compose on Failure
 
-There are cases when you would like the `process-compose` to terminate immediately when one of the processes exits with non `0` exit code. This can be useful when you would like to perform "pre-flight" validation checks on the environment.
+There are cases when you might like `process-compose` to terminate immediately when one of the processes exits with non `0` exit code. This can be useful when you would like to perform "pre-flight" validation checks on the environment.
 
 To achieve that, use `exit_on_failure` restart policy. If defined, `process-compose` will gracefully shut down all the other running processes and exit with the same exit code as the failed process.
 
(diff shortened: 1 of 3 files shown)


adrian-gierakowski:
built and manually tested locally

let me know if you think we should add some tests, and if so, what's the easiest way to run them

F1bonacc1:
Thank you @adrian-gierakowski for the PR!
Would you mind documenting this feature?
BTW, the use case that you mentioned in #77 is very helpful.

Cheers!

adrian-gierakowski:
Sure, will do!

F1bonacc1:
Looks Good, Thanks!

PR #80, opened 2023-07-21
bugfix: fix definition of flake overlays output

adrian-gierakowski:
overlays should not be defined per-system (see: https://discourse.nixos.org/t/how-to-consume-a-eachdefaultsystem-flake-overlay/19420/3)
also, overlay (singular) is not an official output
so we keep overlays.default only

I'm working on a project which doesn't use flakes yet so prefer to consume
via an overlay :)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,26 +7,28 @@
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = { self, nixpkgs, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
+    (flake-utils.lib.eachDefaultSystem (system:
       let
         pkgs = import nixpkgs {
           inherit system;
           overlays = [ self.overlays."${system}".default ];
         };
       in {
-        overlays.default = final: prev: {
-          process-compose = final.callPackage ./default.nix {
-            #version = self.shortRev or "dirty";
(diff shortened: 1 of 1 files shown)


F1bonacc1:
Thanks again @adrian-gierakowski
@thenonameguy was taking care of this code. Probably everything is OK, but if you want another pair of eyes, it's better to be his 😄

thenonameguy (approved):
LGTM, good catch 👍

adrian-gierakowski:
Thanks!

mikeland73:
I think this breaks flake support. 

Previously:

```bash
nix search github:F1bonacc1/process-compose/v0.43.1 --json
{"packages.aarch64-darwin.process-compose":{"description":"A simple and flexible scheduler and orchestrator to manage non-containerized applications","pname":"process-compose","version":"0.43.1"}}
```

But now:

```bash
nix search github:F1bonacc1/process-compose --json
{}
```

`nix profile install github:F1bonacc1/process-compose/v0.43.1`

works fine, but 

```
nix profile install github:F1bonacc1/process-compose
error: cached failure of attribute 'defaultPackage.aarch64-darwin'
```

fails

thenonameguy:
Sorry for the breakage, fixing.

PR #86, opened 2023-08-09
Don't create a new nixpkgs instance when not required, fix overlay attr.

### LINKED ISSUE #85: Reference to non-existing `self.overlays.${system}`
Since #80, it's no longer possible to `nix run` the flake:
```
➜ nix run 'github:f1bonacc1/process-compose' --
error:
       … while evaluating the attribute 'packages.x86_64-linux.process-compose'

         at /nix/store/x13pm9vsklcq0iis25w8k28n8x2w24ml-source/flake.nix:17:36:

           16|       in {
           17|         packages = { inherit (pkgs) process-compose; };
             |                                    ^
           18|         defaultPackage = self.packages."${system}".process-compose;

       … while evaluating a branch condition

         at /nix/store/hbxj731z8hd22lva84px6h9ppl5qjzgx-source/lib/trivial.nix:386:27:

          385|   */
          386|   throwIfNot = cond: msg: if cond then x: x else throw msg;
             |                           ^
          387|

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'x86_64-linux' missing

       at /nix/store/x13pm9vsklcq0iis25w8k28n8x2w24ml-source/flake.nix:14:24:

           13|           inherit system;
           14|           overlays = [ self.overlays."${system}".default ];
             |                        ^
           15|         };
```

R-VdP:
Don't create a new nixpkgs instance when accessing the `packages.process-compose` attribute. We can construct the package from the given nixpkgs input without applying an overlay to it.

This also fixes the issue with the reference to the overlay containing the system.

Fixes #85.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -7,14 +7,16 @@
   inputs.flake-utils.url = "github:numtide/flake-utils";
 
   outputs = { self, nixpkgs, flake-utils }:
+    let
+      mkPackage = pkgs: pkgs.callPackage ./default.nix {
+        #version = self.shortRev or "dirty";
+        date = self.lastModifiedDate;
+        commit = self.shortRev or "dirty";
+      };
+    in
     (flake-utils.lib.eachDefaultSystem (system:
-      let
-        pkgs = import nixpkgs {
-          inherit system;
-          overlays = [ self.overlays."${system}".default ];
-        };
-      in {
-        packages = { inherit (pkgs) process-compose; };
+      {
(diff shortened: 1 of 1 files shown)


thenonameguy (approved):
Verified on `aarch64-darwin` (via: 
`nix run 'github:R-VdP/process-compose/a3df51066bdb6f7c333b30b500666b50a8dded3f'`). 
Thanks for the fix, very elegant and saves the extra nixpkgs instance. 👍

Issue #88, opened 2023-09-02
Allow runnning executables directly without shell wrapper 


adrian-gierakowski:
## Feature Request

I’d like to be able to specify a command as a list of: [executable, args…] and have it ran directly instead of via a shell. 

Note that I’m willing to implement this. 

#### Use Case:

Reducing dependencies needed to run the processes or double-wrapping in a shell, if my executable is already a shell script, which calls other executables. 

#### Proposed Change:

Allow process.command to be an array, in which case it would be invoked directly 

#### Who Benefits From The Change(s)?

People wanting to minimise the deps of their projects.

#### Alternative Approaches

Not sure.

adrian-gierakowski:
seems like union types are a pain in go (I'm might be wrong since I have very little experience with this language) so making command a `string | []string` might not be the best choice 

Adding optional `args []string` and a `bool` flag to disable shell for given process would probably make more sense, although would result in a more verbose config

F1bonacc1:
Hi @adrian-gierakowski,

That's an interesting proposal. Will it work for you to replace the command with `${COMMAND_WITH_ARGS_FOR_SOME_PROC}` environment variable place holder and provide the value for the env variable externally?

adrian-gierakowski:
> That's an interesting proposal. Will it work for you to replace the command with `${COMMAND_WITH_ARGS_FOR_SOME_PROC}`

I think an array filed is required to simplify passing of args with spaces  

I've included a draft implementation in https://github.com/F1bonacc1/process-compose/pull/89

F1bonacc1:
#89 Is merged in v0.75.1

Issue #83, opened 2023-07-24
equivalent of docker-compose run: start and attach to process (stdin\stdout) while running deps in background


adrian-gierakowski:
## Feature Request

I'd like to be able to run a selected process in forground, with stdin\stdout attached, so that I see it's logs and interact with it from command line. All of it's dependent processes should run in the background (no logs output shown in terminal). 

This would be similar to [docker-compose run $SERVICE_NAME](https://docs.docker.com/engine/reference/commandline/compose_run/)

For example, given the following `docker-compose.yaml`:

```yaml
services:
  main:
    image: busybox
    command:
      - sh
      - -c
      - |
        echo running main; while true; read -p "enter some text: " reply && echo $$reply; done
    depends_on:
      - deps

  deps:
    image: busybox
    command:
      - sh
      - -c
      - while true; do date > deps; sleep 1; done
```

when I run `docker-compose run main`, I see the following:

```
[+] Running 1/0
 ⠿ Container example-deps-1  Created                                                                                                                                          0.0s
[+] Running 1/1
 ⠿ Container example-deps-1  Started                                                                                                                                          0.2s
running main
enter some text: 
```

which then allows me to provide input at the prompt

the `deps` container keeps running in the background until the main process exits.

#### Use Case:

I'd like to be able to run tests as the main process and a bunch of deps (databases) as other processes in the background. The main process needs to wait for deps before it starts. I am only interested in seeing the test logs in the terminal, and I'd also like to provide input to the test process while it's running: for example, the [jest](https://jestjs.io/) test runner, when ran in `--watch` mode, can be controlled with various keyboard shortcuts which allow to rerun all tests, or only ones that failed.

#### Proposed Change:

Addition of:
- top level `run` command, taking 1 process name as argument: this would start **all** process defined in the config file, while attaching to the named process
- `run` sub-command to the `process` command, taking 1 process name as argument: this would start (and attach to) only given process. with all processes in it's dependency tree running in the background

a `--no-deps` flag could be nice to have

#### Who Benefits From The Change(s)?

People who'd like to use PC to interactively run some process, while also running other process on which the main one depends

#### Alternative Approaches

One could run deps in PC, while the main process outside of it, and use a wrapper script to start\stop PC as needed.

Also, a workaround I am currently using in CI, when running test process and it's deps via PC (with a config similar to [this](https://github.com/F1bonacc1/process-compose#terminate-process-compose-once-given-process-ends)), is a following bash script:

```bash
[...]

PR #89, opened 2023-09-04
implement process entrypoint and run cmd

adrian-gierakowski:
### process entrypoint

`entrypoint` is an optional array of strings which can be set on the process config
to override default shell (and shellArg) used to run the `command`

When `entrypoint` is set and is non-empty, `command` can also be omitted.

This is useful if one want's to:
1. use a different shell per process
2. run any other executable directly, without a shell wrapper

 ### `run` cmd

Equivalent of [docker-compose run](https://docs.docker.com/engine/reference/commandline/compose_run/)
which can be used to run a chosen process in the foreground (with std(in|out|err) attached),
while it's dependencies are ran in the background (no logs printed to stdio).

Deps can be disabled via --no-deps flag.

Additional arguments can be passed to the process after `--` separator, for example:

```sh
process-compose run main -- arg1 arg2
```

The separator is necessary in order to distinguish process-compose flags from flags
meant to be passed to the process.

While the process is running, the logs of the deps can be inspected using the
usual command:

```
process-compose process logs dep
```

The process being ran will have `availability.exit_on_end` set to `true` and
manually setting it to `false` will have no effect.

implements #88 and #83

### DIFF
--- a/src/app/commander_interface.go
+++ b/src/app/commander_interface.go
@@ -15,4 +15,5 @@ type Commander interface {
 	StderrPipe() (io.ReadCloser, error)
 	Stop(int, bool) error
 	SetCmdArgs()
+	AttachIo()
 }
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -53,6 +53,9 @@ type Process struct {
 	liveProber    *health.Prober
 	readyProber   *health.Prober
 	shellConfig   command.ShellConfig
+	printLogs     bool
+	isMain        bool
+	extraArgs     []string
 }
 
 func NewProcess(
@@ -62,6 +65,9 @@ func NewProcess(
 	processState *types.ProcessState,
 	procLog *pclog.ProcessLogBuffer,
 	shellConfig command.ShellConfig,
+	printLogs bool,
+	isMain bool,
+	extraArgs []string,
 ) *Process {
 	colNumeric := rand.Intn(int(color.FgHiWhite)-int(color.FgHiBlack)) + int(color.FgHiBlack)
 
@@ -78,6 +84,9 @@ func NewProcess(
 		shellConfig:   shellConfig,
 		procStateChan: make(chan string, 1),
 		procReadyChan: make(chan string, 1),
+		printLogs:     printLogs,
+		isMain:        isMain,
+		extraArgs:     extraArgs,
 	}
 
 	proc.procReadyCtx, proc.readyCancelFn = context.WithCancel(context.Background())
@@ -92,7 +101,7 @@ func (p *Process) run() int {
 	}
 
 	if err := p.validateProcess(); err != nil {
-		log.Error().Err(err).Msgf("Failed to run command %s for process %s", p.getCommand(), p.getName())
+		log.Error().Err(err).Msgf(`Failed to run command ["%v"] for process %s`, strings.Join(p.getCommand(), `" "`), p.getName())
 		p.onProcessEnd(types.ProcessStateError)
 		return 1
 	}
@@ -101,7 +110,7 @@ func (p *Process) run() int {
 	for {
 		err := p.setStateAndRun(p.getStartingStateName(), p.getProcessStarter())
 		if err != nil {
-			log.Error().Err(err).Msgf("Failed to run command %s for process %s", p.getCommand(), p.getName())
+			log.Error().Err(err).Msgf(`Failed to run command ["%v"] for process %s`, strings.Join(p.getCommand(), `" "`), p.getName())
 			p.logBuffer.Write(err.Error())
 			p.onProcessEnd(types.ProcessStateError)
 			return 1
@@ -151,16 +160,23 @@ func (p *Process) run() int {
 
 func (p *Process) getProcessStarter() func() error {
 	return func() error {
-		p.command = command.BuildCommandShellArg(p.shellConfig, p.getCommand())
+		p.command = command.BuildCommand(
+			p.procConf.Executable,
+			append(p.procConf.Args, p.extraArgs...),
+		)
 		p.command.SetEnv(p.getProcessEnvironment())
 		p.command.SetDir(p.procConf.WorkingDir)
-		p.command.SetCmdArgs()
-		stdout, _ := p.command.StdoutPipe()
-		stderr, _ := p.command.StderrPipe()
-		go p.handleOutput(stdout, p.handleInfo)
-		go p.handleOutput(stderr, p.handleError)
-		//stdin, _ := p.command.StdinPipe()
-		//go p.handleInput(stdin)
+
+		if p.isMain {
+			p.command.AttachIo()
+		} else {
+			p.command.SetCmdArgs()
+			stdout, _ := p.command.StdoutPipe()
+			stderr, _ := p.command.StderrPipe()
+			go p.handleOutput(stdout, p.handleInfo)
+			go p.handleOutput(stderr, p.handleError)
+		}
+
 		return p.command.Start()
 	}
 }
@@ -346,8 +362,11 @@ func (p *Process) getNameWithSmartReplica() string {
 	return p.procConf.Name
 }
 
-func (p *Process) getCommand() string {
-	return p.procConf.Command
+func (p *Process) getCommand() []string {
+	return append(
+		[]string{(*p.procConf).Executable},
+		append(p.procConf.Args, p.extraArgs...)...,
+	)
 }
 
 func (p *Process) updateProcState() {
@@ -401,13 +420,17 @@ func (p *Process) handleOutput(pipe io.ReadCloser, handler func(message string))
 
 func (p *Process) handleInfo(message string) {
 	p.logger.Info(message, p.getName(), p.procConf.ReplicaNum)
-	fmt.Printf("[%s\t] %s\n", p.procColor(p.getName()), message)
+	if p.printLogs {
+		fmt.Printf("[%s\t] %s\n", p.procColor(p.getName()), message)
+	}
 	p.logBuffer.Write(message)
 }
 
 func (p *Process) handleError(message string) {
 	p.logger.Error(message, p.getName(), p.procConf.ReplicaNum)
-	fmt.Printf("[%s\t] %s\n", p.procColor(p.getName()), p.redColor(message))
+	if p.printLogs {
+		fmt.Printf("[%s\t] %s\n", p.procColor(p.getName()), p.redColor(message))
+	}
 	p.logBuffer.Write(message)
 }
 
--- a/src/app/project_runner.go
+++ b/src/app/project_runner.go
@@ -25,6 +25,8 @@ type ProjectRunner struct {
 	logger           pclog.PcLogger
 	waitGroup        sync.WaitGroup
 	exitCode         int
+	mainProcess      string
+	mainProcessArgs  []string
 }
 
 func (p *ProjectRunner) GetLexicographicProcessNames() ([]string, error) {
@@ -83,7 +85,25 @@ func (p *ProjectRunner) runProcess(config *types.ProcessConfig) {
 		procLog = pclog.NewLogBuffer(0)
 	}
 	procState, _ := p.GetProcessState(config.ReplicaName)
-	process := NewProcess(p.project.Environment, procLogger, config, procState, procLog, *p.project.ShellConfig)
+	isMain := config.Name == p.mainProcess
+	hasMain := p.mainProcess != ""
+	printLogs := !hasMain
+	extraArgs := []string{}
+	if isMain {
+		extraArgs = p.mainProcessArgs
+		config.RestartPolicy.ExitOnEnd = true
+	}
+	process := NewProcess(
+		p.project.Environment,
+		procLogger,
+		config,
+		procState,
+		procLog,
+		*p.project.ShellConfig,
+		printLogs,
+		isMain,
+		extraArgs,
+	)
 	p.addRunningProcess(process)
 	p.waitGroup.Add(1)
 	go func() {
@@ -561,10 +581,18 @@ func (p *ProjectRunner) GetProject() *types.Project {
 	return p.project
 }
 
-func NewProjectRunner(project *types.Project, processesToRun []string, noDeps bool) (*ProjectRunner, error) {
+func NewProjectRunner(
+	project *types.Project,
+	processesToRun []string,
+	noDeps bool,
+	mainProcess string,
+	mainProcessArgs []string,
+) (*ProjectRunner, error) {
 
 	runner := &ProjectRunner{
 		project: project,
+		mainProcess: mainProcess,
+		mainProcessArgs: mainProcessArgs,
 	}
 
 	var err error
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -10,7 +10,7 @@ import (
 // runCmd represents the up command
 var runCmd = &cobra.Command{
 	Use:   "run PROCESS [flags] -- [process_args]",
-	Short: "Run PROCESS in the foreground, and it's dependencies in the background",
+	Short: "Run PROCESS in the foreground, and its dependencies in the background",
 	Long: `Run selected process with std(in|out|err) attached, while other processes run in the background.
 Command line arguments, provided after --, are passed to the PROCESS.`,
 	Args: cobra.MinimumNArgs(1),
--- a/src/cmd/run.go
+++ b/src/cmd/run.go
@@ -1,0 +1,54 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"github.com/f1bonacc1/process-compose/src/api"
+	"github.com/spf13/cobra"
+)
+
+// runCmd represents the up command
+var runCmd = &cobra.Command{
+	Use:   "run PROCESS [flags] -- [process_args]",
+	Short: "Run PROCESS in the foreground, and it's dependencies in the background",
+	Long: `Run selected process with std(in|out|err) attached, while other processes run in the background.
+Command line arguments, provided after --, are passed to the PROCESS.`,
+	Args: cobra.MinimumNArgs(1),
+	// Args: cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		*pcFlags.Headless = false
+
+		processName := args[0]
+
+		if len(args) > 1 {
+			argsLenAtDash := cmd.ArgsLenAtDash()
+			if argsLenAtDash != 1 {
+				message := "Extra positional arguments provided! To pass args to PROCESS, separate them from process-compose arguments with: --"
+				fmt.Println(message)
+				os.Exit(1)
+			}
+			args = args[argsLenAtDash:]
+		} else {
+			// Clease args as they will contain the processName
+			args = []string{}
+		}
+
+		runner := getProjectRunner(
+			[]string{processName},
+			*pcFlags.NoDependencies,
+			processName,
+			args,
+		)
+
+		api.StartHttpServer(false, *pcFlags.PortNum, runner)
+		runProject(runner)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(runCmd)
+
+	runCmd.Flags().BoolVarP(pcFlags.NoDependencies, "no-deps", "", *pcFlags.NoDependencies, "don't start dependent processes")
+	runCmd.Flags().AddFlag(rootCmd.Flags().Lookup("config"))
+
+}
--- a/src/command/command_wrapper.go
+++ b/src/command/command_wrapper.go
@@ -2,6 +2,7 @@ package command
 
 import (
 	"io"
+	"os"
 	"os/exec"
 )
 
@@ -35,6 +36,12 @@ func (c *CmdWrapper) StdoutPipe() (io.ReadCloser, error) {
 
 func (c *CmdWrapper) StderrPipe() (io.ReadCloser, error) {
 	return c.cmd.StderrPipe()
+}
+
+func (c *CmdWrapper) AttachIo() () {
+	c.cmd.Stdin = os.Stdin
+	c.cmd.Stdout = os.Stdout
+	c.cmd.Stderr = os.Stderr
 }
 
 func (c *CmdWrapper) SetEnv(env []string) {
--- a/src/types/process.go
+++ b/src/types/process.go
@@ -15,7 +15,8 @@ type ProcessConfig struct {
 	Name              string
 	Disabled          bool                   `yaml:"disabled,omitempty"`
 	IsDaemon          bool                   `yaml:"is_daemon,omitempty"`
-	Command           string                 `yaml:"command"`
+	Command           *string                `yaml:"command,omitempty"`
+	Entrypoint        *[]string              `yaml:"entrypoint,omitempty"`
 	LogLocation       string                 `yaml:"log_location,omitempty"`
 	LoggerConfig      *LoggerConfig          `yaml:"log_configuration,omitempty"`
 	Environment       Environment            `yaml:"environment,omitempty"`
@@ -31,6 +32,8 @@ type ProcessConfig struct {
 	Extensions        map[string]interface{} `yaml:",inline"`
 	ReplicaNum        int
 	ReplicaName       string
+	Executable        string
+	Args              []string
 }
 
 func (p *ProcessConfig) GetDependencies() []string {
--- a/src/types/process.go
+++ b/src/types/process.go
@@ -15,8 +15,8 @@ type ProcessConfig struct {
 	Name              string
 	Disabled          bool                   `yaml:"disabled,omitempty"`
 	IsDaemon          bool                   `yaml:"is_daemon,omitempty"`
-	Command           *string                `yaml:"command,omitempty"`
-	Entrypoint        *[]string              `yaml:"entrypoint,omitempty"`
+	Command           string                 `yaml:"command"`
+	Entrypoint        []string               `yaml:"entrypoint"`
 	LogLocation       string                 `yaml:"log_location,omitempty"`
 	LoggerConfig      *LoggerConfig          `yaml:"log_configuration,omitempty"`
 	Environment       Environment            `yaml:"environment,omitempty"`
--- a/src/types/validators.go
+++ b/src/types/validators.go
@@ -77,34 +77,28 @@ func (p *Project) assignDefaultProcessValues() {
 
 func (p *Project) validateProcessCommand() {
 	for name, proc := range p.Processes {
-		if proc.Entrypoint == nil {
-			if proc.Command == nil {
-				message := fmt.Sprintf("Error: Neither command nor entrypoint is set for process: %s", name)
-				fmt.Println(message)
-				log.Fatal().Msg(message)
-			}
-			proc.Entrypoint = &[]string{
-				p.ShellConfig.ShellCommand,
-				p.ShellConfig.ShellArgument,
-			}
-		} else {
-			if len(*proc.Entrypoint) == 0 && (proc.Command == nil || *proc.Command == "") {
-				message := fmt.Sprintf("If entrypoint is empty, command needs to be non-empty (procces: %s)", name)
-				fmt.Println(message)
-				log.Fatal().Msg(message)
-			}
-		}
+		command := proc.Command
+		entrypoint := proc.Entrypoint
 
-		if len(*proc.Entrypoint) == 0 {
-			proc.Executable = *proc.Command
-			proc.Args = []string{}
-		} else {
-			entrypoint := *proc.Entrypoint
+		if command != "" {
+			if len(entrypoint) > 0 {
+				message := fmt.Sprintf("Both command and entrypoint are set! Using command and ignoring entrypoint (procces: %s)", name)
+				fmt.Println(message)
+				log.Error().Msg(message)
+			}
+
+			proc.Executable = p.ShellConfig.ShellCommand;
+			proc.Args = []string{
+				p.ShellConfig.ShellArgument,
+				command,
+			}
+		} else if len(entrypoint) > 0 {
 			proc.Executable = entrypoint[0]
 			proc.Args = entrypoint[1:]
-			if proc.Command != nil {
-				proc.Args = append(proc.Args, *proc.Command)
-			}
+		} else {
+			message := fmt.Sprintf("Either command or entrypoint need to be non-empty (procces: %s)", name)
+			fmt.Println(message)
+			log.Fatal().Msg(message)
 		}
 
 		p.Processes[name] = proc
--- a/src/types/validators.go
+++ b/src/types/validators.go
@@ -5,6 +5,7 @@ import (
 	"github.com/f1bonacc1/process-compose/src/command"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"os"
 	"strings"
 )
 
@@ -80,10 +81,10 @@ func (p *Project) validateProcessCommand() {
 		command := proc.Command
 		entrypoint := proc.Entrypoint
 
-		if command != "" {
+		if command != "" || len(entrypoint) == 0 {
 			if len(entrypoint) > 0 {
 				message := fmt.Sprintf("Both command and entrypoint are set! Using command and ignoring entrypoint (procces: %s)", name)
-				fmt.Println(message)
+				fmt.Fprintln(os.Stderr, "process-compose:", message)
 				log.Error().Msg(message)
 			}
 
@@ -92,13 +93,9 @@ func (p *Project) validateProcessCommand() {
 				p.ShellConfig.ShellArgument,
 				command,
 			}
-		} else if len(entrypoint) > 0 {
+		} else {
 			proc.Executable = entrypoint[0]
 			proc.Args = entrypoint[1:]
-		} else {
-			message := fmt.Sprintf("Either command or entrypoint need to be non-empty (procces: %s)", name)
-			fmt.Println(message)
-			log.Fatal().Msg(message)
 		}
 
 		p.Processes[name] = proc
--- a/src/types/validators.go
+++ b/src/types/validators.go
@@ -19,6 +19,7 @@ func (p *Project) ValidateAfterMerge() error {
 	p.assignDefaultProcessValues()
 	p.cloneReplicas()
 	p.copyWorkingDirToProbes()
+	p.validateProcessCommand()
 	return p.validateNoCircularDependencies()
 }
 
@@ -70,6 +71,42 @@ func (p *Project) assignDefaultProcessValues() {
 			proc.Replicas = 1
 		}
 		proc.Name = name
+		p.Processes[name] = proc
+	}
+}
+
+func (p *Project) validateProcessCommand() {
+	for name, proc := range p.Processes {
+		if proc.Entrypoint == nil {
+			if proc.Command == nil {
+				message := fmt.Sprintf("Error: Neither command nor entrypoint is set for process: %s", name)
+				fmt.Println(message)
+				log.Fatal().Msg(message)
+			}
+			proc.Entrypoint = &[]string{
+				p.ShellConfig.ShellCommand,
+				p.ShellConfig.ShellArgument,
+			}
+		} else {
+			if len(*proc.Entrypoint) == 0 && (proc.Command == nil || *proc.Command == "") {
+				message := fmt.Sprintf("If entrypoint is empty, command needs to be non-empty (procces: %s)", name)
+				fmt.Println(message)
+				log.Fatal().Msg(message)
+			}
+		}
+
+		if len(*proc.Entrypoint) == 0 {
+			proc.Executable = *proc.Command
+			proc.Args = []string{}
+		} else {
+			entrypoint := *proc.Entrypoint
+			proc.Executable = entrypoint[0]
+			proc.Args = entrypoint[1:]
+			if proc.Command != nil {
+				proc.Args = append(proc.Args, *proc.Command)
+			}
+		}
+
 		p.Processes[name] = proc
 	}
 }
--- a/flake.nix
+++ b/flake.nix
@@ -14,14 +14,19 @@
         commit = self.shortRev or "dirty";
       };
     in
-    (flake-utils.lib.eachDefaultSystem (system:
-      {
-        packages.process-compose = mkPackage nixpkgs.legacyPackages.${system};
+    (flake-utils.lib.eachDefaultSystem (system: let
+      pkgs = nixpkgs.legacyPackages.${system};
+    in {
+        packages.process-compose = mkPackage pkgs;
         defaultPackage = self.packages."${system}".process-compose;
         apps.process-compose = flake-utils.lib.mkApp {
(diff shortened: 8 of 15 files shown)


adrian-gierakowski:
Happy to extract https://github.com/F1bonacc1/process-compose/issues/88 into separate PR.

regarding https://github.com/F1bonacc1/process-compose/issues/83, the way I made it work is to silence std(out|err) output for all processes apart from the main one passed to `run` as an argument, and the connecting `os.std(in|out|err)` [directly to the main process](https://github.com/F1bonacc1/process-compose/pull/89/files#diff-4771a543439152654dc1c5ae866b7d941e5051ca79380270f23ca31a49a87dc0R45-R48) 

note that I had to skip [p.command.SetCmdArgs()](https://github.com/F1bonacc1/process-compose/pull/89/files#diff-498610a8d4690f9395d51dc6f07f5a8e44c14d5765fcddae7c7f7dc6fba5684eR166) since otherwise things would break in some cases (like running tests with `yarn jest --watch`  in a node project). I guess this means that the main process will receive signals (Ctrl+C) directly. This is ok in my case as I set `exit_on_end` on the main process (which I guess would be a good default anyway).

adrian-gierakowski:
note that I'm not well versed in `go` so don't hesitate to criticise the implementation :)

adrian-gierakowski:
run command can be tested using the config file:

```yaml
shell:
  shell_command: "bash"
  shell_argument: "-c"

processes:
  main:
    command: |
      echo say hi
      read -r in
      echo $in
      echo by
    depends_on:
      dep:
        condition: process_started
    availability:
      exit_on_end: true

  dep:
    command: while true; do echo sleeping... && sleep 2; done
```

run the `main` process:

```sh
./result/bin/process-compose run main -f process-compose.yaml
```

the `dep` will start in background but it's logs will not be output by the command above
instead you can view its logs with:

```sh
./result/bin/process-compose process logs -f dep
```

adrian-gierakowski:
and here's an example of how to configure a process to run without a shell wrapper:

```yaml
processes:
  main:
    no_shell: true
    command: yarn
    args:
      - jest
      - --watch
```

We could potentially default to `no_shell: true` if `args` are set. Technically you could pass extra args to a shell, as in:

```sh
> bash -c 'echo name: $0 arg1: $1' my-name my-arg
name: my-name arg1: my-arg
```

but not sure if that would of use to anyone

F1bonacc1:
> and here's an example of how to configure a process to run without a shell wrapper:
> 
> ```yaml
> processes:
>   main:
>     no_shell: true
>     command: yarn
>     args:
>       - jest
>       - --watch
> ```
> 
> We could potentially default to `no_shell: true` if `args` are set. Technically you could pass extra args to a shell, as in:
> 
> ```shell
> > bash -c 'echo name: $0 arg1: $1' my-name my-arg
> name: my-name arg1: my-arg
> ```
> 
> but not sure if that would of use to anyone

Not sure this is going to be a widely used notation :)
I think that if the args list is not empty, the configured shell can be ignored and run `cmd ...args`

F1bonacc1:
Are there use cases where it only makes sense to run a process as part of the `run` command? In other words, should the interactive process be disabled when ran as part of the full `process-compose.yaml`?

I tested the code (as is) with `vim` and it creates some unwanted visual issues when Tui is disabled.

F1bonacc1:
@adrian-gierakowski 
That's a more than welcome PR!
Having STDIN is a super useful feature, that was asked by other users before.
I was trying to crack the TUI approach, but I like your, more practical and actually working direction, more.

Thank you for that!

adrian-gierakowski:
> I think that if the args list is not empty, the configured shell can be ignored and run cmd ...args

👍 

> Are there use cases where it only makes sense to run a process as part of the run command? In other words, should the interactive process be disabled when ran as part of the full process-compose.yaml?

There might be some, but in my case it works both with `run` and `up`, the latter with or without TUI. With extra args passed from cli, I would setup main process as `yarn jest` and then pass —watch when running with `run` for interactive experience. And if I wanted to do a one shot execution with outputs from all service interleaved I’d use `up`.

adrian-gierakowski:
> I tested the code (as is) with vim and it creates some unwanted visual issues when Tui is disabled.

could you post a minimal config I could use to reproduce? Or a nix expression? Thanks!

F1bonacc1:
> > I tested the code (as is) with vim and it creates some unwanted visual issues when Tui is disabled.
> 
> could you post a minimal config I could use to reproduce? Or a nix expression? Thanks!

```yaml
version: "0.5"
log_level: debug

processes:
  vim:
    command: vim
    args:
      - compose.yaml
    depends_on:
      dep:
        condition: process_started

  dep:
    command: while true; do echo sleeping... && sleep 0.5; done
```

To reproduce:
1. Run: `process-compose -t=false`
2. Exit
3. Run again

I think this can be avoided if interactive processes are disabled when run with others.

adrian-gierakowski:
I guess vim doesn’t have non-interactive mode so it would only make sense to run it via `run`. But many cli tools detect tty and work with and without it.

I think we can leave it up to the user to choose the correct command to make their config work. We could also add a new flag which would disable the service when ran via `up`, but then we’d have to decide if it’s deps should also be disabled.

F1bonacc1:
> I guess vim doesn’t have non-interactive mode so it would only make sense to run it via `run`. But many cli tools detect tty and work with and without it.
> 
> I think we can leave it up to the user to choose the correct command to make their config work. We could also add a new flag which would disable the service when ran via `up`, but then we’d have to decide if it’s deps should also be disabled.

I agree, We can start by letting the users manage the configuration if they want to skip the interactive processes. Since it's an edge case, I think it's a reasonable compromise.
Once we have more feedback and more specific use cases (debuggers, wizards, setup) we can decide on a flag or other behavior.

adrian-gierakowski:
So, do you think adding the capability to pass additional args from cli to the main process ran with `run` cmd would be enough for you to accept this PR? Do you have any comments or improvements suggestions regarding implementation?

F1bonacc1:
> So, do you think adding the capability to pass additional args from cli to the main process ran with `run` cmd would be enough for you to accept this PR? Do you have any comments or improvements suggestions regarding implementation?

Yes, it looks good. There are a few things that I would have done differently, some small typos and no documentation, but for a WIP PR, it's in a great shape :)

adrian-gierakowski:
> Yes, it looks good. There are a few things that I would have done differently, some small typos and no documentation, but for a WIP PR, it's in a great shape :)

Ah yes, I’ll definitely add documentation. I’ll also double check for typos. Happy to make any further changes if you provide specific pointers. Thanks!

src/types/process.go
```diff
@@ -16,6 +16,8 @@ type ProcessConfig struct {
 	Disabled          bool                   `yaml:"disabled,omitempty"`
 	IsDaemon          bool                   `yaml:"is_daemon,omitempty"`
 	Command           string                 `yaml:"command"`
```
F1bonacc1, line 18: Should args and command (with more than one argument) be mutually exclusive?

src/command/command_wrapper.go
```diff
@@ -37,6 +38,16 @@ func (c *CmdWrapper) StderrPipe() (io.ReadCloser, error) {
 	return c.cmd.StderrPipe()
 }
 
+func (c *CmdWrapper) StdinPipe() (io.WriteCloser, error) {
```
F1bonacc1, line 41: Unused

src/app/commander_interface.go
```diff
@@ -13,6 +13,8 @@ type Commander interface {
 	SetDir(string)
 	StdoutPipe() (io.ReadCloser, error)
 	StderrPipe() (io.ReadCloser, error)
+	StdinPipe() (io.WriteCloser, error)
```
F1bonacc1, line 16: Unused

adrian-gierakowski:
@F1bonacc1 apologies for the long silence but I’ve been quite busy after coming back from holidays. I should have some time to work on this in about a week as I’ve got a couple of flights to take 😄

F1bonacc1:
> @F1bonacc1 apologies for the long silence but I’ve been quite busy after coming back from holidays. I should have some time to work on this in about a week as I’ve got a couple of flights to take 😄

@adrian-gierakowski,
Don't worry about it. I know the feeling all too well 😉

src/cmd/run.go
```diff
@@ -0,0 +1,32 @@
...
+		processName := args[0]
+
+		runner := getProjectRunner(args, noDeps, processName)
+		api.StartHttpServer(false, port, runner)
+		runProject(runner)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(runCmd)
```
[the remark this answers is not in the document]
adrian-gierakowski: I would say it’s better to mirror (a subset of) docker-compose’s api to make things intuitive for people familiar with it.

[`docker-compose run`](https://docs.docker.com/engine/reference/commandline/compose_run/) has -i and -t in the sense intended here (which are both true by defraud and hence can be omitted). This is also similar to [`docker run`](https://docs.docker.com/engine/reference/commandline/run/)

[docker-compose up](https://docs.docker.com/engine/reference/commandline/compose_up/) does not have -i and -t has a different meaning (timeout).

src/types/process.go
```diff
@@ -16,6 +16,8 @@ type ProcessConfig struct {
...
+	Args              []string               `yaml:"args,omitempty"`
+	NoShell           bool                   `yaml:"no_shell,omitempty"`
```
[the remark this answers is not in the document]
adrian-gierakowski: Works for me.

src/cmd/run.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"github.com/f1bonacc1/process-compose/src/api"
+	"github.com/spf13/cobra"
+)
+
+// runCmd represents the up command
+var runCmd = &cobra.Command{
+	Use:   "run [PROCESS]",
+	Short: "Run PROCESS in the forground (witn stdin, stdout, stderr attached), and and it's dependencies in the background",
```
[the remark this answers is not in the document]
adrian-gierakowski: 👍

src/types/process.go
adrian-gierakowski: Personally I’d live it to the user. There is an odd chance that an executable has a space in it.

src/types/process.go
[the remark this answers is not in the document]
adrian-gierakowski: Btw. ideally (imho), command would be either a string or an array, and in the latter case we would not use a shell. But I’m go lang fu is not advanced enough to pull this off.

Alternatively we could leave command as it is, and introduce [entrypoint](https://docs.docker.com/compose/compose-file/compose-file-v3/), which would be specified as an array, and be user in place of default shell and shell_argument for given process. This way we’d remain consistent with a subset of docker-compose’s api. 

Thinking about it now, I actually think the latter option would be the best.

src/cmd/run.go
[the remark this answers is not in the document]
F1bonacc1: Sounds good. Let's go with run.

src/types/process.go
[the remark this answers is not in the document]
F1bonacc1: Entrypoint sounds like an excellent compromise.
In docker-compose `entrypoint` is both a string and a slice.

src/types/process.go
F1bonacc1: 👍🏻

src/cmd/run.go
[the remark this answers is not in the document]
thenonameguy: And to be pedantic, "it's" as well 😄 
https://www.merriam-webster.com/grammar/when-to-use-its-vs-its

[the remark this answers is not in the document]
adrian-gierakowski: And `witn` -> `with` :rofl:

adrian-gierakowski:
@F1bonacc1 I believe I now got the desired behaviour implemented. I've updated PR description. 

Still missing documentation, but I'll add this once the implementation is finished (in case you have any further comments you'd like me to address). 

Not sure why tests are failing (haven't ran them before pushing), will have a look at it tomorrow.

adrian-gierakowski:
@F1bonacc1 not quite sure how to run the tests locally. There doesn't seem to be a way to do it via nix. And I also wasn't able to run the command which is used in github workflow from within `nix develop`. Would be great if instructions were added on the Readme about how to setup a dev environment, or at least allow the tests to be ran via `nix flake check`

F1bonacc1:
@adrian-gierakowski,
You are right, I should have made it clearer in the Readme. Especially now when when there are more and more amazing contributions coming our way. Thank you for that! 
If you have `make` you can run `make testrace`.
If not, `go test -race ./src/... ` will do the trick.

adrian-gierakowski:
> @adrian-gierakowski, You are right, I should have made it clearer in the Readme. Especially now when when there are more and more amazing contributions coming our way. Thank you for that! If you have `make` you can run `make testrace`. If not, `go test -race ./src/... ` will do the trick.

no luck:

```
> nix develop
> make testrace
go test -race ./src/...
go: inconsistent vendoring in /home/adrian/code/process-compose:
	github.com/InVisionApp/go-health/v2@v2.1.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/adrg/xdg@v0.4.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/cakturk/go-netstat@v0.0.0-20200220111822-e5b49efee7a5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/f1bonacc1/glippy@v0.0.0-20230203184233-82c6562cecd1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/fatih/color@v1.15.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gdamore/tcell/v2@v2.6.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gin-gonic/gin@v1.9.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gorilla/websocket@v1.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/imdario/mergo@v0.3.16: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/joho/godotenv@v1.5.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/rivo/tview@v0.0.0-20230530133550-8bd761dda819: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/spf13/cobra@v1.7.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/swaggo/swag@v1.16.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	gopkg.in/natefinch/lumberjack.v2@v2.2.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	gopkg.in/yaml.v2@v2.4.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/InVisionApp/go-logger@v1.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/KyleBanks/depth@v1.2.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/bytedance/sonic@v1.9.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/chenzhuoyu/base64x@v0.0.0-20221115062448-fe3a3abad311: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gabriel-vasile/mimetype@v1.4.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gdamore/encoding@v1.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/gin-contrib/sse@v0.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-openapi/jsonpointer@v0.19.6: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-openapi/jsonreference@v0.20.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-openapi/spec@v0.20.9: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-openapi/swag@v0.22.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-playground/locales@v0.14.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-playground/universal-translator@v0.18.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/go-playground/validator/v10@v10.14.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/goccy/go-json@v0.10.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/inconshreveable/mousetrap@v1.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/jezek/xgb@v1.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/josharian/intern@v1.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/json-iterator/go@v1.1.12: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/klauspost/cpuid/v2@v2.2.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/leodido/go-urn@v1.2.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/lucasb-eyer/go-colorful@v1.2.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mailru/easyjson@v0.7.7: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mattn/go-runewidth@v0.0.14: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/modern-go/concurrent@v0.0.0-20180306012644-bacd9c7ef1dd: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/modern-go/reflect2@v1.0.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/pelletier/go-toml/v2@v2.0.8: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/rivo/uniseg@v0.4.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/spf13/pflag@v1.0.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/twitchyliquid64/golang-asm@v0.15.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/ugorji/go/codec@v1.2.11: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/arch@v0.3.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/crypto@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/net@v0.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/term@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/text@v0.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/tools@v0.9.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	google.golang.org/protobuf@v1.30.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	gopkg.in/yaml.v3@v3.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mattn/go-colorable@v0.1.13: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/mattn/go-isatty@v0.0.19: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/rs/zerolog@v1.29.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/swaggo/files@v1.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/swaggo/gin-swagger@v1.6.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	golang.org/x/sys@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
	github.com/InVisionApp/go-health/v2: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
	github.com/cakturk/go-netstat: is replaced in go.mod, but not marked as replaced in vendor/modules.txt

	To ignore the vendor directory, use -mod=readonly or -mod=mod.
	To sync the vendor directory, run:
		go mod vendor
make: *** [Makefile:53: testrace] Error 1
```

F1bonacc1:
> > @adrian-gierakowski, You are right, I should have made it clearer in the Readme. Especially now when when there are more and more amazing contributions coming our way. Thank you for that! If you have `make` you can run `make testrace`. If not, `go test -race ./src/... ` will do the trick.
> 
> no luck:
> 
> ```
> > nix develop
> > make testrace
> go test -race ./src/...
> go: inconsistent vendoring in /home/adrian/code/process-compose:
> 	github.com/InVisionApp/go-health/v2@v2.1.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/adrg/xdg@v0.4.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/cakturk/go-netstat@v0.0.0-20200220111822-e5b49efee7a5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/f1bonacc1/glippy@v0.0.0-20230203184233-82c6562cecd1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/fatih/color@v1.15.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gdamore/tcell/v2@v2.6.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gin-gonic/gin@v1.9.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gorilla/websocket@v1.5.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/imdario/mergo@v0.3.16: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/joho/godotenv@v1.5.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/rivo/tview@v0.0.0-20230530133550-8bd761dda819: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/spf13/cobra@v1.7.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/swaggo/swag@v1.16.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	gopkg.in/natefinch/lumberjack.v2@v2.2.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	gopkg.in/yaml.v2@v2.4.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/InVisionApp/go-logger@v1.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/KyleBanks/depth@v1.2.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/bytedance/sonic@v1.9.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/chenzhuoyu/base64x@v0.0.0-20221115062448-fe3a3abad311: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gabriel-vasile/mimetype@v1.4.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gdamore/encoding@v1.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/gin-contrib/sse@v0.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-openapi/jsonpointer@v0.19.6: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-openapi/jsonreference@v0.20.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-openapi/spec@v0.20.9: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-openapi/swag@v0.22.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-playground/locales@v0.14.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-playground/universal-translator@v0.18.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/go-playground/validator/v10@v10.14.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/goccy/go-json@v0.10.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/inconshreveable/mousetrap@v1.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/jezek/xgb@v1.1.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/josharian/intern@v1.0.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/json-iterator/go@v1.1.12: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/klauspost/cpuid/v2@v2.2.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/leodido/go-urn@v1.2.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/lucasb-eyer/go-colorful@v1.2.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/mailru/easyjson@v0.7.7: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/mattn/go-runewidth@v0.0.14: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/modern-go/concurrent@v0.0.0-20180306012644-bacd9c7ef1dd: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/modern-go/reflect2@v1.0.2: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/pelletier/go-toml/v2@v2.0.8: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/rivo/uniseg@v0.4.4: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/spf13/pflag@v1.0.5: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/twitchyliquid64/golang-asm@v0.15.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/ugorji/go/codec@v1.2.11: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/arch@v0.3.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/crypto@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/net@v0.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/term@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/text@v0.10.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/tools@v0.9.3: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	google.golang.org/protobuf@v1.30.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	gopkg.in/yaml.v3@v3.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/mattn/go-colorable@v0.1.13: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/mattn/go-isatty@v0.0.19: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/rs/zerolog@v1.29.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/swaggo/files@v1.0.1: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/swaggo/gin-swagger@v1.6.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	golang.org/x/sys@v0.9.0: is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt
> 	github.com/InVisionApp/go-health/v2: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
> 	github.com/cakturk/go-netstat: is replaced in go.mod, but not marked as replaced in vendor/modules.txt
> 
> 	To ignore the vendor directory, use -mod=readonly or -mod=mod.
> 	To sync the vendor directory, run:
> 		go mod vendor
> make: *** [Makefile:53: testrace] Error 1
> ```

@adrian-gierakowski,
It looks like you need to get those packages first:
```shell
git clone https://github.com/F1bonacc1/process-compose.git
cd process-compose
go get ./...    # probably the missing step, please notice the 3 dots
make testrace
```

adrian-gierakowski:
@F1bonacc1 tests are fixed now

do let me know if you have any comments, improvement suggestions or if you'd like me to add some tests, thanks!

if you're happy with the implementation (or at least the interface), I'll start working on docs

src/types/process.go
```diff
@@ -15,7 +15,8 @@ type ProcessConfig struct {
 	Name              string
 	Disabled          bool                   `yaml:"disabled,omitempty"`
 	IsDaemon          bool                   `yaml:"is_daemon,omitempty"`
-	Command           string                 `yaml:"command"`
+	Command           *string                `yaml:"command,omitempty"`
+	Entrypoint        *[]string              `yaml:"entrypoint,omitempty"`
```
F1bonacc1, line 19: Why `entrypoint` is a pointer?

src/types/validators.go
```diff
@@ -74,6 +75,42 @@ func (p *Project) assignDefaultProcessValues() {
...
+			}
+		}
+
+		if len(*proc.Entrypoint) == 0 {
+			proc.Executable = *proc.Command
+			proc.Args = []string{}
+		} else {
+			entrypoint := *proc.Entrypoint
+			proc.Executable = entrypoint[0]
+			proc.Args = entrypoint[1:]
```
F1bonacc1, line 104: If `len(*proc.Entrypoint)==1` this code will crash.

F1bonacc1 (changes requested):

src/app/project_runner.go
```diff
@@ -83,7 +85,25 @@ func (p *ProjectRunner) runProcess(config *types.ProcessConfig) {
...
 	}
 	procState, _ := p.GetProcessState(config.ReplicaName)
-	process := NewProcess(p.project.Environment, procLogger, config, procState, procLog, *p.project.ShellConfig)
+	isMain := config.Name == p.mainProcess
+	hasMain := p.mainProcess != ""
+	printLogs := !hasMain
+	extraArgs := []string{}
+	if isMain {
+		extraArgs = p.mainProcessArgs
+		config.RestartPolicy.ExitOnEnd = true
```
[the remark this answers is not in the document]
adrian-gierakowski: since the purpose of `run` is to run a single process and it's deps. Once the main process exits there doesn't seem to be any reason to keep the dependencies around.

src/types/process.go
adrian-gierakowski: since when Entrypont is `!= nil` we disregard `ShellConfig`

https://github.com/F1bonacc1/process-compose/blob/b99cd40adbde9ff32c6aebae0f357671fd091ce8/src/types/validators.go#L98-L107

src/app/project_runner.go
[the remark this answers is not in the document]
F1bonacc1: Don't you think that it should be up to the user to decide that?
Aren't there use cases where the rest of the apps should continue running, even after the tui app exits?

src/types/process.go
F1bonacc1: Yes, I saw that.
Can the same result be achieved if you check that it's not empty?

adrian-gierakowski: The point is that you can set it to empty to execute a command directly, without shell, if the command is simply a name or path of an executable.  

Current implementation allows one to:

1. Use command on it’s own, executed via shell set in ShellConfig. This is the old behaviour
2. Use Entrypoint on it’s own
3. Use Entrypoint with command

In case 2 and 3, ShellConfig is ignored. If Entrypoint happens to be an empty array, the command becomes the executable, since the command (if not nil) is simply appended to the Entrypoint array.

If you think the latter case is not useful, we could indeed default Entrypoint to [] and have it result in behaviour number 1

src/types/validators.go
adrian-gierakowski: hmm, just tested it and it seems to work fine

src/types/process.go
F1bonacc1: I think that cases 1 and 2 are very useful. 3 however, is a bit confusing and doesn't add anything to the functionality that can't be achieved by 1 and 2.
IMHO the flow can be simplified to the following rules:

1. If `command` is not empty (`len(command) > 0`) - run the command. This is the default behavior.
2. If `entrypoint` is not empty - run the `entrypoint` exec `[0]` and args `[1:]`.
3. If both are not empty:
   a. issue an error to log during validation
   b. run the `command`
   c. ignore the `entrypoint`
   d. in strict mode (upcoming) fail the start

What do you think?

src/types/validators.go
F1bonacc1: You are absolutely right!
`entrypoint[1:]` returns an empty slice if `len(*proc.Entrypoint)==1`

src/app/project_runner.go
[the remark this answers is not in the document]
adrian-gierakowski: `run` command doesn't start a TUI, it only works in command line mode. We attach stdio to the main process so that it can be interacted with and we only show the logs for that process. When the process end, we exit PC with the same code as the main process.

Note that you can still attach a TUI to PC which is started via the `run` command, but since interactivity is not implemented in the TUI, starting in TUI mode via `run` would be no different then running `up main-process-name`

Given the above, I think it would be counter intuitive to keep supporting services running once the main process exits as all IO would stop, and the program would simply silently hang there, without passing control back to the shell's command line. 

If you feel strongly about giving users the choice to do, I would suggest keeping current behaviour as default and allow it to be overridden via `--exit-on-end=false`.

But TBH, I would rather wait until somebody actually comes up with a valid use case for it.

src/types/process.go
adrian-gierakowski: Tbh, I don't personally like this pointer business (would very much rather use an Option type, which `go` doesn't seem to provide) so would be happy to use `[]` and `""` signifying "not set" for `Entrypoint` and `command` respectively, if you are happy to drop the use case described above

adrian-gierakowski: I wrote my reply without seeing yours just above :)

src/app/project_runner.go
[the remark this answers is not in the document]
F1bonacc1: CLI Applications like `gdb --tui` is a HUGE added bonus of your PR. I know that it wasn't intended to run TUI, but without your feature it wouldn't be possible in Process Compose 😃

Let's go with the "wait and see" approach that you proposed. If someone will ask for it we can always add another flag (`--exit-on-end=false`)

src/types/process.go
adrian-gierakowski: > 1. If `command` is not empty (`len(command) > 0`) - run the command. This is the default behavior.
> 2. If `entrypoint` is not empty - run the `entrypoint` exec `[0]` and args `[1:]`.
> 3. If both are not empty:
>    a. issue an error to log during validation
>    b. run the `command`
>    c. ignore the `entrypoint`
>    d. in strict mode (upcoming) fail the start
> 

Sounds good. 

Btw. I just refreshed my memory regarding why I implemented it to support `Entrypoint` + `command`: this is how docker and docker-compose work (they also allow both of them to be either a single string or an array). Tbh I always found it to be quite confusing so it's probably not the best model to replicate.

F1bonacc1 (approved):

src/types/validators.go
```diff
@@ -74,6 +75,36 @@ func (p *Project) assignDefaultProcessValues() {
...
+				p.ShellConfig.ShellArgument,
+				command,
+			}
+		} else if len(entrypoint) > 0 {
+			proc.Executable = entrypoint[0]
+			proc.Args = entrypoint[1:]
+		} else {
+			message := fmt.Sprintf("Either command or entrypoint need to be non-empty (procces: %s)", name)
+			fmt.Println(message)
+			log.Fatal().Msg(message)
```
adrian-gierakowski, line 101: just realised that this is a breaking change, since previously empty command was allowed

F1bonacc1:
Everything looks great @adrian-gierakowski!
I will wait for the documentation before merging.
I will need to move it to the new location that I plan to publish after the merge.
Spoiler: https://f1bonacc1.github.io/

src/types/validators.go
F1bonacc1: You are right, ignore the empty command for now. I will fix it in the Strict mode.

src/app/process.go
```diff
@@ -151,16 +160,23 @@ func (p *Process) run() int {
 
 func (p *Process) getProcessStarter() func() error {
 	return func() error {
-		p.command = command.BuildCommandShellArg(p.shellConfig, p.getCommand())
+		p.command = command.BuildCommand(
+			p.procConf.Executable,
+			append(p.procConf.Args, p.extraArgs...),
```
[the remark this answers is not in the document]
adrian-gierakowski: Could you give an example of the final command you’d like executed?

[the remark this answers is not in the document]
F1bonacc1:
```yaml
processes:
  ls:
    command: "ls"
```
```shell
process-compose run ls -- /tmp
```
`run ls` will ignore the extraArg `/tmp`
It worked well with `entrypoint` instead of `command`.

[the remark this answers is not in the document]
adrian-gierakowski: This prints: `args: 0 1 2`

```bash
bash -c 'echo args: $0 $1 $2' 0 1 2 
```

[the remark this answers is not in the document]
F1bonacc1: Try:
```shell
bash -c "ls" /tmp
```
vs.
```shell
bash -c "ls /tmp"
```

PR #118, opened 2023-12-11
imp: add nix ci gha job

johnalotoski:
* Add basic nix build gha ci to ensure the process-compose package builds before new releases are tagged.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -19,7 +19,7 @@ buildGoModule rec {
   nativeBuildInputs = [ installShellFiles ];
 
   vendorHash = "sha256-0On/Rg8c9g45qbLuwhP/ZIGosu0X1uzXfAoddgTCDkg=";
-  # vendorHash = lib.fakeSha256;
+  # vendorHash = lib.fakeHash;
 
   postInstall = ''
     mv $out/bin/{src,process-compose}
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,18 @@
+name: Nix
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+jobs:
+  build:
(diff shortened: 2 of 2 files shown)


default.nix
```diff
@@ -19,7 +19,7 @@ buildGoModule rec {
   nativeBuildInputs = [ installShellFiles ];
 
   vendorHash = "sha256-0On/Rg8c9g45qbLuwhP/ZIGosu0X1uzXfAoddgTCDkg=";
```
F1bonacc1, line 21: sha256-mfErA+LnafVuvIfecZCOgw3DZp/if1xlM8UaaaPuj0M=

F1bonacc1 (changes requested):

F1bonacc1 (approved):
👍🏻

default.nix
johnalotoski: Hi @F1bonacc1, I don't think we want to change that hash just yet.

It builds fine with the existing hash for me on nixpkgs 23.05 and 23.11 on a couple different nix versions 2.17.x and 2.19.x.  I'm looking into why it's reporting a hash change when it runs in GHA.  I wasn't expecting that.

I'd like to investigate a little more as I have time over the next few days.

Apologies in advance for creating some failed GHA attempt history as I investigate.

F1bonacc1: The reason it failed is that I accepted a `dependabot` commit just a few minutes before your PR.
It appears that this commit breaks a foreground capability of PC.
I reverted the commit and now the GHA passes without any issues.

johnalotoski: Ah!  That confused me! :).  Ok, I'll revert back to the origin nix version I had in the ci job with one more force push, and then I think it will be good to go.

F1bonacc1 (approved):

PR #122, opened 2023-12-22
Expose method and expected status code on http checks

gkze:


src/health/probe.go
```diff
@@ -2,13 +2,14 @@ package health
...
 	"fmt"
+	"net/http"
 	"net/url"
 	"strings"
 )
 
 type Probe struct {
 	Exec             *ExecProbe `yaml:"exec,omitempty"`
-	HttpGet          *HttpProbe `yaml:"http_get,omitempty"`
+	Http             *HttpProbe `yaml:"http,omitempty"`
```
F1bonacc1, line 12: This breaks backward compatibility.
Do you think we can have both `http` and `http_get` fields at the same time?
Even nicer would be to use the deprecation warning:
https://github.com/F1bonacc1/process-compose/blob/main/src/types/deprecation.go

src/health/probe_test.go
```diff
@@ -71,7 +71,7 @@ func TestProbe_validateAndSetDefaults(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			p := &Probe{
 				Exec:             tt.fields.Exec,
-				HttpGet:          tt.fields.HttpGet,
+				Http:             tt.fields.Http,
```
F1bonacc1, line 74: Can you please add a test to validate that the `Method` and the `StatusCode` are set correctly?

src/loader/mutators.go
```diff
@@ -139,9 +139,9 @@ func renderProbe(probe *health.Probe, tpl *templater.Templater, vars types.Vars)
...
 	if probe.Exec != nil {
 		probe.Exec.Command = tpl.RenderWithExtraVars(probe.Exec.Command, vars)
-	} else if probe.HttpGet != nil {
-		probe.HttpGet.Path = tpl.RenderWithExtraVars(probe.HttpGet.Path, vars)
-		probe.HttpGet.Host = tpl.RenderWithExtraVars(probe.HttpGet.Host, vars)
-		probe.HttpGet.Scheme = tpl.RenderWithExtraVars(probe.HttpGet.Scheme, vars)
+	} else if probe.Http != nil {
+		probe.Http.Path = tpl.RenderWithExtraVars(probe.Http.Path, vars)
+		probe.Http.Host = tpl.RenderWithExtraVars(probe.Http.Host, vars)
+		probe.Http.Scheme = tpl.RenderWithExtraVars(probe.Http.Scheme, vars)
```
F1bonacc1, line 145: Can you please add the `Method` rendering here?

F1bonacc1 (changes requested):
Hey @gkze,

Thanks for your PR!!
Look really good!

There are a few minor requests:
1. The transition from `http_get` to `http` breaks backward compatibility. Can we support both? We can say that it will be deprecated later with a code that I shared with you in the comment.
2. Can you please add tests that validate the new default values `GET` and `200`.
3. Can you please add the Method to the renderer?
4. I think this new and shiny functionality deserves documentation.

Cheers!

src/health/probe.go
gkze: Thanks, will do. Can you give an example of how this is done?

src/health/probe_test.go
gkze: ⚡

src/loader/mutators.go
gkze: ⚡

src/health/probe.go
F1bonacc1: Certainly.
Check out this commit as an example:
https://github.com/F1bonacc1/process-compose/commit/45fceb3abfdf36e8205d2322806ea16be23d6a5e

I mistakenly used `on-failure` instead of `on_failure`, once discovered I kept both of them:
1. I issued a warning during the first month after deprecation.
2. Error + 5s sleep during the second month
3. Exit 1 after that.

I would go ahead and add the `deprecationCheck` to `validators.go` and assign the `http_get` values (if exist) to `http` (if not defined) in `mutators.go`.

If that's too many steps, let me know and I will add this content.

F1bonacc1: IMO in case both of them are defined you can fail the loading in `loader.validators.go`.
Please let me know if you need any help with this feature.

dzmitry-lahoda:
> This branch has conflicts that must be resolved

@gkze would it be possible to find some time slot to handle conflicts? thank you

gkze:
Sorry for the long delay here. Yeah I can update this PR sometime by the end of this week.

gkze:
Just thinking about this - I realize my fork has fallen far enough behind that it might be worth restarting the change. I see that status code was already added.

For `.ValidateAndSetDefaults()`, if we add an `Http` field alongside the `HttpGet` field it would be possible to have a case where both are set. There is currently no error being returned from that method. Should the signature change to check that if both are set to return an error stating that those two fields are mutually exclusive?

PR #135, opened 2024-02-04
move setInputCapture to mainGrid, except ctrl+c

stijnveenman:
Currently, when you open the find Dialog, pressing escape to close it does not work. First.
First, i tried checking if a `DialogPage` existed, and if it does, close it. I added this in the `ActionLogFindExit`, but this closing of the dialog page is not always connected to `Esc` is the keybinding is changed. So it needed an extra check for that. This didn't feel right to me. 

Thinking about it, it felt odd the escape never reached the PageDialog even though that page has focus. It seems that's because all the keybindings are captures in the `appView`. This always goes first, before any active page. 

So I moved this mapping to the `mainGrid` page. This caused Control+C to not be captures. So i moved just this keybinding so the appView. 

This allows the PageDialog to handle Escape before it's send to the mainGrid. Which allows the search dialog to close itself on escape. 

I tried to make sure all keybindings still work, and as far as i can tell they do.

### DIFF
--- a/src/tui/view.go
+++ b/src/tui/view.go
@@ -106,7 +106,9 @@ func newPcView(project app.IProject) *pcView {
 	pv.pages = tview.NewPages().
 		AddPage(PageMain, pv.mainGrid, true, true)
 
+	pv.mainGrid.SetInputCapture(pv.onMainGridKey)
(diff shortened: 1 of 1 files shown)


F1bonacc1 (approved):
Looks good. Thanks!

Issue #133, opened 2024-01-30
Toggle visibility of disabled services


jankramer:
## Feature Request

It would be great if we could toggle the visibility of disabled services at run-time.

#### Use Case:

I love the feature introduced in #123. However, in one of the projects I use process-compose, there are around 200 processes. Of these services, typically only a handful are actually running. Before v0.77.8, the process overview would be very clear, but now with all the disabled processes it is less readable.

#### Proposed Change:

If a user would be able to toggle the visibility of disabled processes, this would allow the overview to remain clear during normal use, and if we want to start an additional process that wasn't started yet, you can easily use the toggle to find and start the process, without having to restart every service.

#### Who Benefits From The Change(s)?

Users with many processes that typically only run a small subset.

#### Alternative Approaches
Sorting the process list by PID does group the processes that are currently running, but that prevents users from sorting by any other column.

F1bonacc1:
Hi @jankramer,

There some additional alternatives that should be considered:
1. Using namespaces there is an existing toggle to show/hide namespaces.
2. Using the `-d` flag to start without the disabled processes (not at a toggle).
3. Using the `-n` flag to start only a specified list (`-n NS1 -n NS2 ...`) of namespaces.

If none of that works for your use case. What shortcut would you recommend for the proposed toggle?

jankramer:
Hi @F1bonacc1,

Thanks for the suggestions!

1.  Namespaces are not the best fit for this use case because processes are used across multiple domains, but can only be in 1 namespace in process-compose (understandably!).
2. Starting with the `-d` flag would indeed solve the readability issue, but then we can no longer quickly start a disabled process without having to restart process-compose. I was actually quite happy that this is now possible since v0.77.8.
3. The `-n` flag suffers from the same limitation as mentioned in 1.

If the shortcut `CTRL+D` is still free, that would be my recommendation.

F1bonacc1:
Added in v0.85.0.

jankramer:
Awesome, just tried it out, and it works great. Thanks for your efforts!

PR #136, opened 2024-02-04
Feature: Filter on process list

stijnveenman:
Add the ability to filter on process list using `/` (vim keybinding). 
Can be confirmed with enter (has the same issue as https://github.com/F1bonacc1/process-compose/pull/135)
Or cancelled with Escape. On start it resets the search, and searches as you type to see the result. 
If cancelled early it will reset the search. 

https://github.com/F1bonacc1/process-compose/assets/126677968/83be52e7-7a94-4bd8-ab8d-5866c2f566f7

I did run into one issue where, if the search results in an empty list, it hangs the process. The processes are still running in the background (seen by the logs updating). But the TUI doesn't respond anymore, i haven't been able to figure out what the reasoning is behind this. If i click escape quickly (to cancel the search), the problem doesn't show up.

### DIFF
--- a/src/tui/proc-filter.go
+++ b/src/tui/proc-filter.go
@@ -1,0 +1,46 @@
+package tui
+
+import (
+	"github.com/gdamore/tcell/v2"
+	"github.com/rivo/tview"
+)
+
+func (pv *pcView) showProcFilter() {
+	const fieldWidth = 50
+	f := tview.NewForm()
+	f.SetCancelFunc(func() {
+		pv.pages.RemovePage(PageDialog)
+	})
+	f.SetItemPadding(1)
+	f.SetBorder(true)
+	f.SetFieldBackgroundColor(tcell.ColorLightSkyBlue)
+	f.SetFieldTextColor(tcell.ColorBlack)
+	f.SetButtonsAlign(tview.AlignCenter)
+	f.SetTitle("Search Process")
+	f.AddInputField("Search For", pv.logsText.getSearchTerm(), fieldWidth, nil, nil)
+	f.AddCheckbox("Case Sensitive", false, nil)
+	f.AddCheckbox("Regex", false, nil)
+	searchFunc := func() {
+		pv.pages.RemovePage(PageDialog)
+	}
+
+	f.AddButton("Search", searchFunc)
+	f.AddButton("Cancel", func() {
+		pv.pages.RemovePage(PageDialog)
+	})
+	f.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+		switch event.Key() {
+		case tcell.KeyEnter:
+			searchFunc()
+		case tcell.KeyEsc:
+			pv.pages.RemovePage(PageDialog)
+		default:
+			return event
+		}
+		return nil
+	})
+	f.SetFocus(0)
+	// Display and focus the dialog
+	pv.pages.AddPage(PageDialog, createDialogPage(f, fieldWidth+20, 11), true, true)
+	pv.appView.SetFocus(f)
+}
--- a/src/tui/proc-filter.go
+++ b/src/tui/proc-filter.go
@@ -33,6 +33,15 @@ func (pv *pcView) showProcFilter() {
 		searchFunc()
 	})
 
+	f.GetFormItem(0).(*tview.InputField).SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+		if event.Key() == tcell.KeyEnter {
+			pv.pages.RemovePage(PageDialog)
+			return nil
+		}
+
+		return event
+	})
+
 	f.AddButton("Search", func() {
 		pv.pages.RemovePage(PageDialog)
 	})
@@ -43,8 +52,6 @@ func (pv *pcView) showProcFilter() {
 
 	f.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
 		switch event.Key() {
-		case tcell.KeyEnter:
-			pv.pages.RemovePage(PageDialog)
 		case tcell.KeyEsc:
 			pv.resetProcessSearch()
 			pv.pages.RemovePage(PageDialog)
@@ -54,6 +61,7 @@ func (pv *pcView) showProcFilter() {
 		return nil
 	})
 	f.SetFocus(0)
+	pv.resetProcessSearch()
 	// Display and focus the dialog
 	pv.pages.AddPage(PageDialog, createDialogPage(f, fieldWidth+20, 11), true, true)
 	pv.appView.SetFocus(f)
--- a/src/tui/proc-filter.go
+++ b/src/tui/proc-filter.go
@@ -17,22 +17,36 @@ func (pv *pcView) showProcFilter() {
 	f.SetFieldTextColor(tcell.ColorBlack)
 	f.SetButtonsAlign(tview.AlignCenter)
 	f.SetTitle("Search Process")
+
 	f.AddInputField("Search For", pv.logsText.getSearchTerm(), fieldWidth, nil, nil)
 	f.AddCheckbox("Case Sensitive", false, nil)
 	f.AddCheckbox("Regex", false, nil)
-	searchFunc := func() {
-		pv.pages.RemovePage(PageDialog)
-	}
 
-	f.AddButton("Search", searchFunc)
-	f.AddButton("Cancel", func() {
+	searchFunc := func() {
+		searchTerm := f.GetFormItem(0).(*tview.InputField).GetText()
+		caseSensitive := f.GetFormItem(1).(*tview.Checkbox).IsChecked()
+		isRegex := f.GetFormItem(2).(*tview.Checkbox).IsChecked()
+
+		pv.searchProcess(searchTerm, caseSensitive, isRegex)
+	}
+	f.GetFormItem(0).(*tview.InputField).SetChangedFunc(func(_ string) {
+		searchFunc()
+	})
+
+	f.AddButton("Search", func() {
 		pv.pages.RemovePage(PageDialog)
 	})
+	f.AddButton("Cancel", func() {
+		pv.resetProcessSearch()
+		pv.pages.RemovePage(PageDialog)
+	})
+
 	f.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
 		switch event.Key() {
 		case tcell.KeyEnter:
-			searchFunc()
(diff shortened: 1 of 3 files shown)


src/tui/proc-filter.go
```diff
@@ -0,0 +1,68 @@
...
+
+	searchFunc := func() {
+		searchTerm := f.GetFormItem(0).(*tview.InputField).GetText()
+		caseSensitive := f.GetFormItem(1).(*tview.Checkbox).IsChecked()
+		isRegex := f.GetFormItem(2).(*tview.Checkbox).IsChecked()
+
+		pv.searchProcess(searchTerm, isRegex, caseSensitive)
+	}
+	f.GetFormItem(0).(*tview.InputField).SetChangedFunc(func(_ string) {
+		searchFunc()
```
F1bonacc1, line 33: Can you pass the text to the `searchFunc` here?
```golang
searchFunc := func(searchTerm string) {
  caseSensitive := f.GetFormItem(1).(*tview.Checkbox).IsChecked()
  isRegex := f.GetFormItem(2).(*tview.Checkbox).IsChecked()
  
  pv.searchProcess(searchTerm, isRegex, caseSensitive)
}
f.GetFormItem(0).(*tview.InputField).SetChangedFunc(func(text string) {
  searchFunc(text)
})

File: src/tui/proc-filter.go [hunk 2]
```diff
@@ -0,0 +1,68 @@
...
+		pv.pages.RemovePage(PageDialog)
+	})
+	f.SetItemPadding(1)
+	f.SetBorder(true)
+	f.SetFieldBackgroundColor(tcell.ColorLightSkyBlue)
+	f.SetFieldTextColor(tcell.ColorBlack)
+	f.SetButtonsAlign(tview.AlignCenter)
+	f.SetTitle("Search Process")
+
+	f.AddInputField("Search For", pv.logsText.getSearchTerm(), fieldWidth, nil, nil)
```
F1bonacc1, line 21:
```golang
pv.logsText.getSearchTerm()
```
Looks like a copy-paste from the log search 😉
It should be the current search term if wasn't reset.

src/tui/proc-table.go
```diff
@@ -46,6 +47,12 @@ func (pv *pcView) fillTableData() {
 			pv.procTable.RemoveRow(row)
 			continue
 		}
+
+		if pv.processRegex != nil && !pv.processRegex.MatchString(state.Name) {
```
F1bonacc1, line 51: I think `pv.processRegex` should be protected with a mutex.
It's being read and written from 2 different threads.

F1bonacc1 (changes requested):
This looks great. Users with many processes will ask themself how did they live without this feature!

A few small changes and we are good to go.

stijnveenman:
I am one of said users with many processes, which is why i wanted this feature 😄 

I've fixed the remarks. And added a mutex. But it doesn't seem to fix the issue. The issue also doesn't happen during the search (when the regex is being written). But happens only if you submit the search with an empty result set. Which makes me suspect it's got something to do with the log window updating and doing something with the selected process, when no process is selected

src/tui/proc-table.go
```diff
@@ -209,6 +221,51 @@ func (pv *pcView) setTableSorter(sortBy ColumnID) {
...
+	defer pv.processRegexMtx.Unlock()
+
+	pv.processRegex = nil
+	go pv.appView.QueueUpdateDraw(func() {
+		pv.fillTableData()
+	})
+}
+
+func (pv *pcView) searchProcess(search string, isRegex, caseSensitive bool) error {
+	pv.processRegexMtx.Lock()
```
F1bonacc1, line 242: That's a good start that you added a mutex to the regex assignment, but the use of the `pv.processRegex` is still unprotected during match.
What you'll probably want to do is implement 2 funcs:

```golang
func (pv *pcView) setProcRegex(reg *regexp.Regexp) {
	pv.processRegexMtx.Lock()
	defer pv.processRegexMtx.Unlock()
	//it can be nil when you want to reset
	pv.processRegex = reg
}

func (pv *pcView) matchProcRegex(procName string) bool {
	pv.processRegexMtx.Lock()
	defer pv.processRegexMtx.Unlock()
	if pv.processRegex != nil {
		return pv.processRegex.MatchString(procName)
	}
	return false
}
```

stijnveenman: That makes sense, i do wonder. Does this need a mutex? 
The `proc-filter` form creates a new regex for every input, and saves the pointer to `pv.processRegex`.
When the `fillTableData` gets the current pointer to the current regex. If another search input it done during the `fillTableData` call, a new regex is created and written to `pv`. `fillTableData` still points to the old regex though, and this regex is not garbage collected. So at worst its slightly outdated?

src/tui/proc-table.go
```diff
@@ -46,13 +47,24 @@ func (pv *pcView) fillTableData() {
...
 		if state.IsRunning {
 			runningProcCount += 1
 		}
 		row += 1
 	}
+	if row == 1 {
+		pv.procTable.SetSelectable(false, false)
+	} else {
+		pv.procTable.SetSelectable(true, false)
+	}
```
F1bonacc1, line 67: I think you encountered my bug, which I already fixed in my env.
Instead of this block, can you try:
```golang
if pv.procTable.GetRowCount()-1 > row - 1{
```
Notice the `row - 1`
In the line below.

src/tui/proc-table.go
F1bonacc1: 2 different threads writing to the same memory location, even for a new value assignment, is a recipe for disaster :)

src/tui/proc-table.go
stijnveenman: I added this block as a workaround for the issue in `tview`. It seems to break if the currently selected row is removed from the table. Which i was also able to replicate in isolation. 

At one point i tried to fix it by calling `table.clear()` before rendering the rows, which means it didn't have to clean up the remaining rows. And it still had the same issue (with the lines you mentioned removed). 

I tried it with this change and run into the same issue.  (to test it)
- enter a filter that results in no processes
- press j or k

F1bonacc1: That's strange. I couldn't replicate it in my environment by, for example, hiding a selected namespace `Ctrl+G`.
Even if it's the last row, leaving me with 0 rows (except the title row).

stijnveenman: Interesting. How did you filter all rows using namespaces? Doesn't every namespace in the list have at least one project?

stijnveenman: Weird. If i
- use the process filter to filter everything but nginx. 
- filter away nginx using namespace filter
- press j
i run into the same issue (notice it only crashes when i actually click j)

F1bonacc1: You are right...
Once I click `j` the same happens to me.

F1bonacc1: It doesn't crash though, for me the process compose is stuck and consumes a lot of CPU.

stijnveenman: Yes sorry, that's what i ment by "crashing" , actually the background processes also keep running fine. (As seen by the logs updating by looking at the logfile directly)

F1bonacc1: I attached a goroutine profiler. It looks like the table is stuck in this loop:
https://github.com/rivo/tview/blob/861aa94d61c899b32a1304d61b840d0178ece408/table.go#L1344

```
#	0x65ae14	github.com/rivo/tview.(*Table).InputHandler.func1.2+0x54			/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/table.go:1375
#	0x65aca3	github.com/rivo/tview.(*Table).InputHandler.func1.5+0xa3			/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/table.go:1440
#	0x65a5a4	github.com/rivo/tview.(*Table).InputHandler.func1+0x4e4				/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/table.go:1598
#	0x65a08f	github.com/rivo/tview.(*Table).InputHandler.(*Box).WrapInputHandler.func2+0x4f	/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/box.go:167
#	0x63bb16	github.com/rivo/tview.(*Grid).InputHandler.func1+0x1b6				/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/grid.go:277
#	0x63b92f	github.com/rivo/tview.(*Grid).InputHandler.(*Box).WrapInputHandler.func2+0x4f	/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/box.go:167
#	0x650801	github.com/rivo/tview.(*Pages).InputHandler.func1+0xa1				/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/pages.go:311
#	0x65072f	github.com/rivo/tview.(*Pages).InputHandler.(*Box).WrapInputHandler.func2+0x4f	/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/box.go:167
#	0x627737	github.com/rivo/tview.(*Application).Run+0x557					/home/eugene/go/pkg/mod/github.com/rivo/tview@v0.0.0-20240101144852-b3bd1aa5e9f2/application.go:344
#	0xb3a3fd	github.com/f1bonacc1/process-compose/src/tui.SetupTui+0x13d			/home/eugene/projects/go/process-compose/src/tui/view.go:474
#	0xb7ebb8	github.com/f1bonacc1/process-compose/src/cmd.startTui+0x178			/home/eugene/projects/go/process-compose/src/cmd/project_runner.go:92
```

stijnveenman: Given this remaining issue, is there anything else in this PR that still needs to be addressed? 
Do you think this workaround should be used for now? GIven that #133 will have a similar issue? Or wait for an upstream fix?

F1bonacc1: I already merged your PR locally.
I believe I will release a new version over the weekend.
Does this timeline work for you?

stijnveenman: Awesome! no rush, i was just wondering if there were any more changes needed 👌

PR #147, opened 2024-03-02
feat: shutdown in reverse dependency order

### LINKED ISSUE #145: process-compose doesn't terminate in dependency order
given process B which depends on process A, I would expect the shutdown routine to be somewhat like:

```
1. shutdown B.
2. wait for the shutdown command to run
3. shutdown A
```

instead, this is what I see in the logs:

```
24-03-01 21:18:46.686 DBG terminating zookeeper with timeout 10 ...
24-03-01 21:18:47.749 DBG terminating kafka with timeout 10 ...
24-03-01 21:18:52.871 INF Project completed
```

even though, in my case, kafka depends on zookeeper, and its shutdown command has `sleep 5`. I would expect the zookeeper shutdown routine to only finish after those 5 seconds.

- you can see the 5 second sleep time between "terminating kafka" and "project completed"
- I would have expected "terminating kafka" to print first, then "terminating zookeeper" printing only after 5 seconds

anmonteiro:
processes that depend on others to start may also need to finish before them and perform some cleanup. This change makes sure that's the case.

fixes #145

### DIFF
--- a/fixtures-code/process-compose-shutdown-inorder.yaml
+++ b/fixtures-code/process-compose-shutdown-inorder.yaml
@@ -1,0 +1,29 @@
+version: "0.5"
+
+log_level: debug
+log_length: 1000
+
+processes:
+  procA:
+    command: |
+      trap 'echo "A: exit"' SIGTERM
+      echo "A: starting"
+      sleep 3
+
+  procB:
+    command: |
+      trap 'echo "B: exit"' SIGTERM
+      echo "B: starting"
+      sleep 3
+    depends_on:
+      procA:
+        condition: process_started
+
+  procC:
+    command: |
+      trap 'echo "C: exit"' SIGTERM
+      echo "C: starting"
+      sleep 3
+    depends_on:
+      procB:
+        condition: process_started
--- a/src/app/system_test.go
+++ b/src/app/system_test.go
@@ -1,10 +1,14 @@
 package app
 
 import (
+	"bufio"
 	"github.com/f1bonacc1/process-compose/src/loader"
+	"github.com/f1bonacc1/process-compose/src/types"
 	"os"
 	"path/filepath"
 	"reflect"
+	"slices"
+	"strings"
 	"testing"
 	"time"
 )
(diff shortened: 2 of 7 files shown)


F1bonacc1 (changes requested):
Hey @anmonteiro,

Thanks for the PR, there are definitely scenarios where this feature is required.
I am a bit concerned that for cases where this is not required, this will make the shutdown sequence much longer.

Since it's a behavior change I don't think we can make it the new default.
A global `is_ordered_shutdown` flag (default `false`) will do the trick.

Will you be able to add a few `ShutDownProject()` tests to `system_test.go` to make sure that it works as expected?

P.S.
Loved the algorithm you used to wait for the dependencies' completion without shutting them down. Nice!

anmonteiro:
@F1bonacc1 thanks for the early draft review. I believed I addressed all your review comments in the latest commit.

Just missing the tests, which I'm working on right now.

anmonteiro:
added a skeleton of a test case shutting down the runner and checking for 0 running processes, but I'm not sure how to check that the processes are actually stopped in the order we expect.

fixtures-code/process-compose-shutdown-inorder.yaml
```diff
@@ -0,0 +1,31 @@
...
+      procA:
+        condition: process_started
+
+  procC:
+    command: |
+      echo "C: starting"
+    isDaemon: true
+    shutdown:
+      command: |
+        echo "C: exiting"
```
thenonameguy, line 28: You could write these to a file, then assert at the end for the contents to be ordered backwards.
```suggestion
        echo "C: exiting" >> shutdown-inorder.log
        sleep 1
```

F1bonacc1: Another approach:
https://github.com/F1bonacc1/process-compose/commit/9ac79177aeed4fb471faa644fbe4ccf9ad9a7360

anmonteiro: Nice, thanks. I cherry-picked it to this branch.

anmonteiro:
Not sure why it fails in CI, `make test` succeeds locally.

src/app/system_test.go
```diff
@@ -406,3 +410,93 @@ func TestSystem_TestProcListToRun(t *testing.T) {
...
+		}
+
+		time.Sleep(10 * time.Millisecond)
+		err = runner.ShutDownProject()
+		if err != nil {
+			t.Errorf(err.Error())
+			return
+		}
+		runningProcesses := 0
+		for _, proc := range runner.runningProcesses {
```
F1bonacc1, line 473: Race condition here.
You can replace this block with `runner.GetProcessesState()` (plural)

F1bonacc1:
Race condition

> Not sure why it fails in CI, `make test` succeeds locally.

Race condition on `runner.runningProcesses`.
Try `make testrace`
Left you a comment at the proper line.

src/app/system_test.go
F1bonacc1: Fixed here:
https://github.com/F1bonacc1/process-compose/commit/27c3ab97973b78499a38d50fbd41d1d4514224eb

anmonteiro:
I actually believe it might be the log rendering using ANSI escape codes but not the wanted string slice.

anmonteiro:
pushed a fix using `GetProcessesState()` and adding `NoColor: true` to the logger configuration.

F1bonacc1:
Looks good! Thanks!

F1bonacc1:
Will you be able to squash?

anmonteiro:
> Will you be able to squash?

Absolutely. Done.

PR #148, opened 2024-03-02
fix: wait for processes to start when `condition: process_started`

anmonteiro:
I noticed that, in my use case (`A(depends_on: B started) -> B(depends_on: C healthy) -> C`):

- B successfully waits for C to be healthy
- A starts at the same time as C, when it should start after B

this change fixes the problematic behavior above.


F1bonacc1:
Great!
Thanks!

PR #182, opened 2024-05-05
support a few more styling options

cdmistman:
some of the styling options weren't changing as i'd expected them to. i added 2 new styling fields `contrastBgColor` and `attentionBgColor` and assigned them to the theme styling's `ContrastBgColor` and `MoreContrastBgColor` fields, respectively.

I also updated all of the existing themes' options to reflect their existing values before this change. I couldn't find any elements that changed colors with these patched fields


F1bonacc1:
That's great @cdmistman!

Thanks for the PR!
In which situations those colors are used?
I fetched the PR, but I can't find the difference 😄

cdmistman:
the components i was trying to style:
- the "quit" dialog (`contrastBgColor`)
- the background colors in the other dialogs (ie in the search dialog - also `contrastBgColor`)
- the `Shutting down...` text (afaict the only element that uses the new `attentionBgColor`)

cdmistman:
> I fetched the PR, but I can't find the difference 😄

that's what i was hoping! opt-in styling changes that nobody will notice but silently expands more theming options :)

might be good for somebody to also update the builtin themes with better-chosen values for these options, but in the meantime as long as nothing changes i think it's ok :)

should also be api-compatible on the patch level (ie, `1.2.1` release) since user themes inherit from the default theme

F1bonacc1:
Got it, thanks.
That looks great.

Agree with your second point. Now that the theming engine is here, given how easy it is to add and edit themes, the idea is to have at least as many themes as KDE Konsole 😄

PR #188, opened 2024-05-22
feat: Watchman integration

cdmistman:
been putting this together, i forked https://github.com/sjansen/watchman for this lmao

primary motivation is as an analog to the [Develop spec](https://docs.docker.com/compose/compose-file/develop/) in the docker-compose spec.

at work, we use process-compose for our dev workflow, and the only part of my workflow that's missing is filesystem watching. while this could be implemented as a shell script, i think it would be great if process-compose directly integrated with watchman for this feature so as not to eat up precious inotify allotments (we do a lot of virtualization).

happy to accept feedback - i know that no issue was opened, nor has there been any discussion of such a feature. it's a little project i gave myself for my own sake as i'd like to use process-compose in a personal project :)


cdmistman:
i know it's not working yet, i also want to clean up the watchman library i forked and publish a proper release. figured i'd open this to see if there was interest in the first place for such a feature

cdmistman:
> What do you mean by this? Could you explain how adding this functionality to PC directly, vs wrapping individual programs which need to restart on file changes with tools like https://github.com/watchexec/watchexec, would help with that? Thanks!

watchman follows a server-client model. the user's filesystem is split into a collection of "projects" (user-configurable, usually means a directory with one of `.git`, `.svn`, `.hg`, or `.watchmanconfig`). when a client wishes to listen for changes to that directory, it must
1. establish an ipc connection to the watchman server
2. request that the project containing that directory be watched
3. subscribe to the directory within the project being watched

what this means is that the inotify allotments are only ever held by the watchman server. when the server receives an inotify notification, it filters the paths that are irrelevant to a given subscription and sends any relevant paths in a subscription ipc message. effectively, this is inotify in userspace :)

cdmistman:
after saying all of that, i think the way i set up subscriptions for each process wrong. instead of using a single watchman subscription for all processes, i should use a single subscription per process that needs it. then i don't need to do further filtering in the client (like i do in https://github.com/F1bonacc1/process-compose/pull/188/commits/6653b14f147ccbfdcedc9ea752e303b7d5099874#diff-defe37158035c547155c5ae19d82fc87e70e6ccbe0d61b261c201872f92b2d00R125-R185)

go.mod
```diff
@@ -31,8 +31,10 @@ replace github.com/cakturk/go-netstat => github.com/f1bonacc1/netstat v0.0.0-202
 require (
 	github.com/InVisionApp/go-logger v1.0.1 // indirect
 	github.com/KyleBanks/depth v1.2.1 // indirect
+	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/bytedance/sonic v1.11.6 // indirect
 	github.com/bytedance/sonic/loader v0.1.1 // indirect
+	github.com/cdmistman/watchman v0.2.1-0.20240521013409-16b1ae353832 // indirect
```
cdmistman, line 37: this won't be pinned to a revision by the time i mark this pr as ready. once i finish this integration and clean up the library's api and do some more tests i'll cut a 0.3 release

src/app/process.go
```diff
@@ -457,8 +490,8 @@ func (p *Process) updateProcState() {
 		p.procState.Name = p.getName()
 	}
 	p.procState.IsRunning = isRunning
-
 }
+
 func (p *Process) setStartTime(startTime time.Time) {
```
cdmistman, line 495: i should use this value to reset the watchman subscription clock

src/app/watchman.go
```diff
@@ -0,0 +1,185 @@
...
+			return
+		}
+
+		notif, ok := rawNotif.(*protocol.Subscription)
+		if !ok {
+			log.Debug().Msg("got a non-subscription notification, ignoring")
+		}
+
+		if notif.IsFreshInstance() {
+			log.Debug().Msg("fresh instance notification, ignoring")
```
cdmistman, line 94: oops forgot continues

src/app/watchman.go
```diff
@@ -0,0 +1,185 @@
...
+		files := []string{}
+		for _, rawFile := range notif.Files() {
+			// if the query fields change, this will need to be a map[string]any
+			file := rawFile.(string)
+			files = append(files, file)
+		}
+
+		for _, sub := range subs {
+			sub.updates <- files
+		}
```
cdmistman, line 106: should change this to send to the channel associated with a subscription name. the name should be something like `process-compose-dev-{process config name}`

src/app/watchman.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return &WatchmanSub{ch, config}
+}
+
+type WatchmanSub struct {
+	updates chan []string
+	config  *types.Watch
+}
+
+func (s *WatchmanSub) Recv() ([]string, bool) {
```
cdmistman, line 125: should change this to only receive on the channel. data associated with a subscription should be tracked by watchman, not in the layer adapting the client

src/types/process.go
```diff
@@ -2,16 +2,26 @@ package types
...
+	PlaceHolderValue = "-"
+)
 
 type Processes map[string]ProcessConfig
+
 type Environment []string
+
+type Watch struct {
+	Glob   string   `yaml:"path"`
+	Ignore []string `yaml:"ignore"`
```
cdmistman, line 22: not *quite* the same as docker-compose's. docker-compose uses just a `Path` where i'm using a `Glob` - i was thinking it might be more convenient to do a glob, but a path might make more sense?

dzmitry-lahoda:
Our use case for that is next

```nix
    settings.processes.rollman = mkIf cfg.rollman.enable {
      command = pkgs.writeShellApplication {
        name = "rollman";
        runtimeInputs = [ config.rustPackage ];
        text = ''
          set -x
          cargo run -p rollman --bin=nord-dev -- start \
            --data-dir ${cfg.rollman.dataDir} \
            --eth-rpc http://localhost:8545 \
            --contract-addr "$(cat /tmp/deloy-testnet-proxy)" \
            --genesis-block 1
        '';
      };
      namespace = "nord";
      depends_on."deploy-local-testnet".condition = "process_completed_successfully";
      depends_on."rollman-genesis".condition = "process_completed_successfully"; 
    };
```

So we have processes which start just building rust codebases. 
I would be interesting to watch files and restart, so getting new builds.

So questions is how watch would interact with

https://f1bonacc1.github.io/process-compose/launcher/  

https://f1bonacc1.github.io/process-compose/health/   (in general, and Auto Restart if not Healthy specifically

For example, when restarting, what is live/ready state during and after? What should dependency processes do (cascading)? Should feat just copy docker-compose behavior for that?

PR #201, opened 2024-07-12
[Fixes #197] Gracefully shutdown HTTP server

### LINKED ISSUE #197: Race condition while shutting down process-compose
## Defect

Make sure that these boxes are checked before submitting your issue -- thank you!

- [x] Included the relevant configuration snippet
- [x] Included the relevant process-compose log (log location: `process-compose info`)
- [x] Included a [Minimal, Complete, and Verifiable example] (https://stackoverflow.com/help/mcve)

#### Version of `process-compose`:
Verified on main branch and v1.6.1

#### OS environment:
macOS, Linux

#### Steps or code to reproduce the issue:
Every now and then (seemingly at random) I would see this error when calling `process-compose down`:
```
24-07-11 14:59:45.839 FTL failed to stop project error="Post \"http://unix/project/stop\": EOF"
```
I thought it was probably me misconfiguring something, but I've come to realize it's a race condition in this function in `pc_api.go`:
```go
func (api *PcApi) ShutDownProject(c *gin.Context) {
	api.project.ShutDownProject()
	c.JSON(http.StatusOK, gin.H{"status": "stopped"})
}
```
I suspect there is a race between shutting down the processes and actually sending the response.

Since this is a race condition I suspect you run into it more often when you have processes that are very quick to shut down. This is the case in the scenarios I've encountered the issue:
```yaml
# service-config.yaml
processes:
  sleeping1:
    command: "sleep 999999"
  sleeping2:
    command: "sleep 999999"
```

In order to reproduce this error you can run `process-compose` under a debugger in one terminal and make calls from the client in another terminal.
```text
# term1
$ dlv debug github.com/f1bonacc1/process-compose/src -- -f service-config.yaml -u pc.sock --tui=false
...
> break pc_api.go:270
> c
```
The debugger will now pause while `process-compose` waits for API calls.
```text
# term2
$ process-compose down -u pc.sock
```
`term2` will now hang (I suppose it's waiting for a response) and `term1` will proceed until it hits the breakpoint. If you go back to `term1` and kill `process-compose` by leaving the debugger, you'll see the exact error I mentioned above.
```text
# term1
> quit
```
```text
# term2
24-07-11 14:59:45.839 FTL failed to stop project error="Post \"http://unix/project/stop\": EOF"
```

#### Expected result:
`process-compose` should reliably and gracefully shutdown when requested.

#### Actual result:
In some cases, `process-compose` does not shut down cleanly.

dcarley:
To prevent a race condition seen in https://github.com/F1bonacc1/process-compose/issues/197:

1. On the client:
  a. User runs `process-compose down`
  b. Client makes request to `/project/stop`
2. On the server:
  a. `ProjectRunner.ShutDownProject()` stops all processes
  b. `cmd.runProject()` stops blocking
  c. Exits before `PcApi.ShutDownProject` has written the response
3. On the client:
  a. Gets `EOF` reading the response because the server has gone away

I can't think of a way to reproduce this in a test, except for something
convoluted like creating a custom client that introduces a delay in
reading the response.

It can be "manually" simulated by adding a small delay here though:

```patch
diff --git a/src/api/pc_api.go b/src/api/pc_api.go
index d123257..d069399 100644
--- a/src/api/pc_api.go
+++ b/src/api/pc_api.go
@@ -3,6 +3,7 @@ package api
 import (
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/f1bonacc1/process-compose/src/app"
 	"github.com/gin-gonic/gin"
@@ -268,6 +269,7 @@ func (api *PcApi) GetProcessPorts(c *gin.Context) {
 // @Router /project/stop [post]
 func (api *PcApi) ShutDownProject(c *gin.Context) {
 	api.project.ShutDownProject()
+	time.Sleep(10 * time.Millisecond)
 	c.JSON(http.StatusOK, gin.H{"status": "stopped"})
 }
```

And using this config:

```yaml
processes:
  one:
    command: sleep infinity
  two:
    command: sleep infinity
```

Before this change:

```console
bash-5.2$ go run src/main.go up --config services.yaml --tui=false &
[1] 8171
bash-5.2$ go run src/main.go down
24-07-12 15:38:40.332 FTL failed to stop project error="Post \"http://localhost:8080/project/stop\": EOF"
exit status 1
[1]+  Done                    go run src/main.go up --config services.yaml --tui=false
```

After this change:

```console
bash-5.2$ go run src/main.go up --config services.yaml --tui=false &
[1] 8432
bash-5.2$ go run src/main.go down
```

### DIFF
--- a/src/api/server.go
+++ b/src/api/server.go
@@ -5,34 +5,56 @@ import (
 	"github.com/f1bonacc1/process-compose/src/app"
 	"github.com/gin-gonic/gin"
 	"github.com/rs/zerolog/log"
+	"net"
+	"net/http"
 	"os"
 )
 
 const EnvDebugMode = "PC_DEBUG_MODE"
 
-func StartHttpServerWithUnixSocket(useLogger bool, unixSocket string, project app.IProject) {
+func StartHttpServerWithUnixSocket(useLogger bool, unixSocket string, project app.IProject) (*http.Server, error) {
 	router := getRouter(useLogger, project)
 	log.Info().Msgf("start UDS http server listening %s", unixSocket)
+
+	os.Remove(unixSocket)
+	server := &http.Server{
+		Handler: router.Handler(),
+	}
+
+	listener, err := net.Listen("unix", unixSocket)
+	if err != nil {
+		return server, err
+	}
+
 	go func() {
-		os.Remove(unixSocket)
-		err := router.RunUnix(unixSocket)
-		if err != nil {
+		defer listener.Close()
+		defer os.Remove(unixSocket)
+
+		if err := server.Serve(listener); err != nil && err != http.ErrServerClosed {
 			log.Fatal().Err(err).Msgf("start UDS http server on %s failed", unixSocket)
 		}
 	}()
+
+	return server, nil
 }
 
-func StartHttpServerWithTCP(useLogger bool, port int, project app.IProject) {
+func StartHttpServerWithTCP(useLogger bool, port int, project app.IProject) (*http.Server, error) {
 	router := getRouter(useLogger, project)
 	endPoint := fmt.Sprintf(":%d", port)
 	log.Info().Msgf("start http server listening %s", endPoint)
+
+	server := &http.Server{
+		Addr:    endPoint,
+		Handler: router.Handler(),
+	}
+
 	go func() {
-		err := router.Run(endPoint)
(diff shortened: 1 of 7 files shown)


F1bonacc1 (approved):
That looks really good!
Thanks for fixing this so quickly!

PR #205, opened 2024-07-23
feat: Add memory column

### LINKED ISSUE #204: Ability to see process memory footprint
## Feature Request
I would like to be able to see the memory footprint of running processes.

#### Use Case:
I want to use process-compose to assist local development of microservices. Sometimes I need several non-dockerized services running at the same time, and seeing memory footprint of individual processes would be a big help to me.

#### Proposed Change:
Add a memory column in the TUI

People running long processes like APIs with a considerable memory footprint.

#### Alternative Approaches

themartdev:
I think psutil is a useful dependency to add, it opens up a lot of doors for cross-platform process instrumentation. We could potentially add think like CPU percentage and such. (although I'm not sure if it would make sense to squeeze CPU% in there)

Fixes #204


src/app/process.go
```diff
@@ -471,6 +474,23 @@ func (p *Process) getStartTime() time.Time {
...
+		return 0
+	}
+	proc, err := process.NewProcess(int32(p.procState.Pid))
+	if err != nil {
+		log.Err(err).Msgf("Could not find process")
+		return -1
+	}
+	meminfo, err := proc.MemoryInfo()
+	if err != nil {
+		log.Err(err).Msgf("Error retrieving memory stats for process: %d", p.procState.Pid)
```
F1bonacc1, line 488: Can you please log the process name?
```golang
log.Err(err).
Str("process", p.getName()).
Int("pid", p.procState.Pid).
Msg("Error retrieving memory stats for process")
```

F1bonacc1 (commented):
Looks great!
A few small comments.
Don't worry about the CI failing. `default.nix` will require a `vendorHash` update because of the new package.

F1bonacc1 (approved):
Great! Thanks!

themartdev:
How do you fix the nix build? I'm not familiar with it :thinking:

F1bonacc1:
> How do you fix the nix build? I'm not familiar with it 🤔

I will take care of that once I merge.

F1bonacc1:
Squashed and Merged, thanks again!!! ❤️

PR #216, opened 2024-08-07
Print log write errors to stderr rather than stdout

ysndr:
Currently the write error message emitted by `process-compose process logs --tail` is indistinguishable from actual log output coming from the process.

```yaml
processes:
  never_exit:
    command: sleep infinity
  foo:
    command: |
      echo 1
      echo 2
      echo 3
```

will log as:

```
$ process-compose process logs -u test.sock -n 2 foo 2>/dev/null
2
3
write close: write unix ->test.sock: write: broken pipe
```

This change will make it so that `write close: write unix ->test.sock: write: broken pipe` would be printed to stderr, i.e. the same command as above would log as:

```
$ process-compose process logs -u test.sock -n 2 foo 2>/dev/null
2
3
```

### DIFF
--- a/src/client/logs.go
+++ b/src/client/logs.go
@@ -8,6 +8,7 @@ import (
 	"github.com/rs/zerolog/log"
 	"io"
 	"net"
+    "os"
 	"sync/atomic"
 )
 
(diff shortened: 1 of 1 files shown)


F1bonacc1 (approved):
LGTM
Thanks

PR #228, opened 2024-08-19
fix: Prefer os.Process.Signal to syscall.Kill

anrddh:
This protects us from a potential (but rare) race condition where

1. the process we are trying to signal terminates of its own accord
2. its PID gets reaped in the `os.Process.Wait` call
3. a new process with the same PID gets spawned

before our signal actually gets sent.

This is achieved by using the higher-level `os.Process` API directly, which
has some synchronization to prevent `os.Process.Wait` and `os.Process.Signal`
from racing.

### DIFF
--- a/src/command/stopper_unix.go
+++ b/src/command/stopper_unix.go
@@ -28,7 +28,7 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
 		Msg("Stop Unix process.")
 
 	if parentOnly {
-		return syscall.Kill(c.Pid(), syscall.Signal(sig))
+		return c.cmd.Process.Signal(syscall.Signal(sig))
 	}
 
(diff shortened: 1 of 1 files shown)


F1bonacc1 (approved):
Looks great!
Thanks.
How did you manage to catch it?!? 😄

anrddh:
I've encountered lots of subtle bugs around `fork`, `exec`, `kill`, `wait` and friends before, so I usually read code that uses them very carefully :)

It's fairly hard to use those syscalls correctly. I've even found discussion for this sort of thing in the Go standard library:
https://github.com/golang/go/commit/cea29c4a358004d84d8711a07628c2f856b381e8
https://github.com/golang/go/issues/9382
https://github.com/golang/go/issues/13987

PR #240, opened 2024-08-30
Forcefully terminate all processes with a configured shutdown timeout

### LINKED ISSUE #214: Force terminate apps if shutdown.timeout_seconds is defined, without requiring shutdown.command to be defined
#### Use Case:

I drive a number of processes that don't reliably shutdown on `SIGTERM`. Usually, this is down to buggy behavior on the part of those processes. It would be nice if Process Compose was robust to such processes, and provided a way to kill them reliably.

#### Proposed Change:

If `shutdown.timeout_seconds` is defined for a process, then `SIGKILL` it after that many seconds, if it hasn't shutdown by then.

Anyone driving processes that don't shutdown gracefully reliably.

#### Alternative Approaches

We can use `pkill -9` in `shutdown.command` to get by for now, but this feels like a bit of a hack. It requires care to ensure processes running outside Process Compose with similar names don't get accidentally killed.

anrddh:
Fixes #214 

I'm not entirely convinced this change is race free, but it should be. I would appreciate if someone double checked my handling of the `p.waitDone` channel.

I assume that a modification to the docs is in order. I can take a look at that once this change is otherwise accepted.

### DIFF
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -73,6 +73,8 @@ type Process struct {
 	stdin               io.WriteCloser
 	passProvided        bool
 	isTuiEnabled        bool
+	canceller           context.CancelFunc
+	waitDone            chan struct{}
 }
 
 func NewProcess(opts ...ProcOpts) *Process {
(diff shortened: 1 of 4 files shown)


F1bonacc1:
Resolved as part of https://github.com/F1bonacc1/process-compose/commit/891f742a6366180ea30d0671e759fc4a961c82d5

PR #246, opened 2024-09-17
tui: don't log.Fatal on appView.Run error

gcurtis:
When the user closes their terminal (as opposed to ctrl-c), the terminal might close the tty without waiting for process-compose to exit. This causes `pcv.appView.Run` to return an error, which triggers a `log.Fatal`:

```
ERR Failed to start TUI error="read /dev/tty: input/output error"
```

This leads to a race where `log.Fatal` might call `os.Exit` before the project's processes can be stopped, leaving them orphaned.

Change `log.Fatal` to `log.Error` and explicitly shutdown the project to make sure processes aren't left behind.

### DIFF
--- a/src/tui/view.go
+++ b/src/tui/view.go
@@ -542,7 +542,8 @@ func RunTUIAsync(project app.IProject, options ...Option) {
 	setupTui(project, options...)
 	go func() {
 		if err := pcv.appView.Run(); err != nil {
-			log.Fatal().Err(err).Msgf("Failed to start TUI")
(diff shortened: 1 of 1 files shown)


gcurtis:
Hey @F1bonacc1, this is something we ran into when running `devbox services up` (which runs the process-compose TUI) in the VS Code terminal. Let me know if this fix makes sense. It seems to eliminate the race, but it's also a bit tricky to test.

F1bonacc1 (approved):
Thanks!
Good catch!

Issue #294, opened 2025-01-07
project X is already specified in files to load


ryantm:
## Defect

When you specify 3 process-compose yaml files and there are extends involved it errors with "project X is already specified in files to load". (If you specify just 2, it works fine.)

```
[ryantm@replit1:/tmp/process-compose]$ tree
.
├── a
│   └── process-compose.yaml
├── b
│   └── process-compose.yaml
├── process-compose-a.yaml
├── process-compose-b.yaml
└── process-compose.yaml

3 directories, 5 files

[ryantm@replit1:/tmp/process-compose]$ nix run nixpkgs#process-compose -- version
Process Compose
Version:        v1.46.0
Commit:         6a1799e
Date (UTC):     2024-12-20T22:09:30Z
License:        Apache-2.0
Discord:        https://discord.gg/S4xgmRSHdC

Written by Eugene Berger

[ryantm@replit1:/tmp/process-compose]$ nix run nixpkgs#process-compose -- -f process-compose.yaml -f process-compose-a.yaml -f process-compose-b.yaml
25-01-06 19:05:21.351 FTL error="project a/process-compose.yaml is already specified in files to load"

[ryantm@replit1:/tmp/process-compose]$ bat process-compose.yaml
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: process-compose.yaml
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ version: "0.5"
   2   │
   3   │ processes:
   4   │   top-hi:
   5   │     command: echo top-hi && sleep 200
───────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

[ryantm@replit1:/tmp/process-compose]$ bat process-compose-a.yaml
───────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: process-compose-a.yaml
───────┼─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ version: "0.5"
   2   │ extends: "./a/process-compose.yaml"
[...]

PR #295, opened 2025-01-07
fix: stop loading the same project multiple times 294

ryantm:
Why
===
* https://github.com/F1bonacc1/process-compose/issues/294

What changed
===
* Make a copy of the FileNames and loop over that instead, so when loadExtendProject adds to the FileNames, it doesn't try to load it again

Test plan
===
* The repro in issue 294 now passes


F1bonacc1 (approved):
Good catch!
Thanks Ryan.

PR #304, opened 2025-02-07
feat: up support detached-with-tui

mwain:
6b9ce00 added `detached-with-tui` for the root command. This adds support to the up


F1bonacc1 (approved):
Great!
Thanks Michael!

PR #306, opened 2025-02-10
feat: support namespace arg in project update command

mwain:
**Why?**
Enable loading/unloading processes by namespace.

This allows grouping processes under namespaces, making it easier to manage dependencies in different environments. For example:
- A **dev** namespace can be used to start only development dependencies.
- The **default** namespace can be used for full integration testing.

By supporting `--namespace`, users can update projects more granularly, starting/stopping only the required processes.

### DIFF
--- a/process-compose.yaml
+++ b/process-compose.yaml
@@ -97,6 +97,23 @@ processes:
       success_threshold: 1
       failure_threshold: 3
 
+  process5:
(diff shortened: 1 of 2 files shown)


F1bonacc1 (approved):
Great!
Thanks Michael!

PR #308, opened 2025-03-17
Clarify readiness vs. liveness probe

### LINKED ISSUE #298: purpose of liveness_probe unclear
Hi,

Given that the `liveness_probe` doesn't trigger restarts, I wonder what its purpose is. I'm using a `ready_log_line` to mark a process as "ready", and `readiness_probe`s serves as a health check to restart processes in case of error, but the purpose of `liveness_probe`s is unclear to me.

The summary of the kubernetes docs is clear:

- Startup Probe — Checks if the container has successfully started.
- Liveness Probe — Checks if the container is still running. 
- Readiness Probe — Checks if the container is ready to accept traffic.

Should the liveness probe not invoke a restart when the restart policy is `always`?

Related issue: When using a `ready_log_line` to detect 'readiness,' I can't also use a `readiness_proble` to run health checks on interval. The logline is a much faster way to detect when a service is up, but given that the `liveness_probe` doesn't trigger restarts, those processes have no way of being restarted on failure.

In addition; when the liveness check fails, the process is still listed as "Running" as if all was fine.

9999years:
Readiness probes are used for processes, liveness probes are used for daemons.

Closes #298.

### DIFF
--- a/www/docs/health.md
+++ b/www/docs/health.md
@@ -6,6 +6,8 @@ Probes configuration and functionality are designed to work similarly to [Kubern
 
 ## Liveness Probe
 
+Liveness probes are used to monitor a [**daemon's**](launcher.md#background-detached-processes) state. If unhealthy, the daemon will transition to the next defined [`availability`](launcher.md#auto-restart-on-exit) state. If it's `always` this will cause a restart, if none the daemon will transition to the completed or failed state.
+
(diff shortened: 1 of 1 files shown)


F1bonacc1 (approved):
Awesome!
Thanks!

PR #310, opened 2025-03-17
Don't say "Shutting Down" when detaching

### LINKED ISSUE #309: "Shutting Down" message shown when detaching
When detaching a `process-compose` client, the "Shutting Down" message is used. This implies that the project is being shut down, even though it's not. The message should be clearer instead.

9999years:
If `process-compose` is detaching from a remote project, it's not shutting down the processes, it's just detaching from _viewing_ them.

Closes #309.


F1bonacc1 (approved):
Looks good.
Thanks!

PR #314, opened 2025-03-18
Add `--shortcuts`, `$PC_SHORTCUTS_FILES`

### LINKED ISSUE #313: Shortcuts are unintuitive
Most of the shortcuts for actions I want to take regularly are on F-keys, so I just have to memorize that (e.g.) `F10` is detach/exit. I'd prefer to have mnemonic shortcuts, like `q` for quit.

The table columns do have mnemonic shortcuts, but those are used much less frequently. With a few processes running, it's usually not necessary to sort the table at all.

A couple ideas here:

1. The default shortcuts could be changed. This is somewhat disruptive, but as long as the shortcuts are shown in the UI, I think it would be reasonable.
2. We could have multiple shortcuts for each action, allowing mnemonic shortcuts without displacing the existing default shortcuts. Unfortunately, this would require changing the architecture of how shortcuts are represented and configuration files are parsed.
3. Perhaps the shortcuts for sorting the table could be moved behind a prefix, e.g. `s` and then `h` to sort by health (currently bound to just `h`).
4. Add a `--shortcuts-config` option to pass extra shortcuts configuration files, which are separate from `--config` files.

Open to suggestions on this, feel free to close if you think the current shortcuts are good enough.

9999years:
In addition to `~/.config/process-compose/shortcuts.yaml`, load shorcuts configuration files from a new `--shortcuts` command-line argument and a new `$PC_SHORTCUTS_FILES` environment variable.

This lets projects ship their own shortcuts configuration files in addition to user-level configuration files.

Partially closes #313.


F1bonacc1 (commented):
Looks good, Rebecca!

2 small requests:
1. I think copilot's comment makes sense. It looks like the Env Var was supposed to be before CLI. It will help to document it as well.
2. I use auto-generated docs for CLI commands. Usually it means running `make docs` and committing the changes.

Thank you!

9999years:
OK, thanks! I've updated the docs and added `--shortcuts` to `process-compose up` as well.

PR #321, opened 2025-04-04
Shutdown instantly

9999years:
Previously, attempting to shut down `process-compose` would incur a 1-second delay.

With #312, this meant that `process-compose --detached-with-tui --detach-on-success` would take longer than 1 second to return, even if the project was already running and healthy (and therefore the command should have been a no-op).

Removing this sleep makes that command take ~90ms instead of ~1050ms.


F1bonacc1 (approved):
I still don't know how do I feel about removing this "Shutting Down..." outro, but I don't want to get in the way of progress 😄

Issue #177, opened 2024-04-18
Unable to override `disabled` option


Invizory:
## Defect

[“Adding and overriding configuration” section in the documentation](https://f1bonacc1.github.io/process-compose/merge/#adding-and-overriding-configuration) states:

> For single-value options like `command`, `working_dir` or `disabled`, the new value replaces the old value.

However, the old value of `disabled` is not actually replaced if the new value is `false`.

#### Version of `process-compose`:

1.2.0

#### OS environment:

Ubuntu 22.04.4 LTS

#### Steps or code to reproduce the issue:

`process-compose.yml`:

```yml
version: "0.5"
processes:
  test:
    disabled: true
    command: echo foo
```

`process-compose.override.yml`:

```yml
processes:
  test:
    disabled: false
    command: echo bar
```

Run `process-compose run test`.

#### Expected result:

```
bar
```

#### Actual result:

```
```

F1bonacc1:
Hi @Invizory,

This is the expected behavior. 
See: https://github.com/darccio/mergo/issues/24#issuecomment-194487379

raszi:
I understand that the underlying library works like that, but then how can I re-enable a service that was disabled in an override file?

F1bonacc1:
I see that some additional comments were added to that thread.
I will reopen this issue, maybe something can be done here after all.

Another temporary workaround can be using vars in your command:
https://github.com/F1bonacc1/process-compose/blob/78d512551927148e34f37d86c2fc371f0efac08a/process-compose.override.yaml#L173-L177

raszi:
In the end, we succeeded with the env var substitution as @thenonameguy wrote, but we needed to disable the strict mode because a disabled service was complaining about another disabled dependency of it.

F1bonacc1:
For backward compatibility, the new `is_disabled` (string) attribute was introduced in v1.63.0. The new behavior is documented here:
https://f1bonacc1.github.io/process-compose/merge/?h=is_di#how-to-use-is_disabled

PR #326, opened 2025-04-09
Add `--enable`, `--disable` override arguments

9999years:
Add `--enable` (`$PC_ENABLE`) and `--disable` (`$PC_DISABLE`) arguments which allow overriding whether or not a process is run.

This can be used to implement optional processes. For example, the following process will not be run by default:

    processes:
      hello:
          command: echo 'Hello World from Process Compose'
          disabled: true

But a user who wants to enable that functionality can set `PC_ENABLE=hello` when starting `process-compose` to enable the process.

Related: https://github.com/F1bonacc1/process-compose/issues/177


9999years:
Needs tests still.

F1bonacc1:
Hey @9999years,

I think this can be useful for people who don't want to edit the configuration file to try new things.
I have concerns with this functionality when reloading the `compose.yaml` (`CTRL+L`) or when calling `process-compose project update another-process-compose.yaml`.
This will enable/disable all the processes according to configuration and ignore user's selection in CLI.

This is a result of passing `opts` with CLI input to the `mutator` functions. It means that the `project` is not the only source of truth for project mutations.

I am interested to hear your opinion about this?

9999years:
Interesting, thanks! I'll see if I can figure out how this interacts with the `process-compose.yaml` reloading feature (we're not using that feature so I didn't think to test it with this change).

9999years:
I think I'm going to close this for now and just use `disabled: $SOME_ENV_VAR` to disable/enable processes. Might be worth revisiting in the future though.

Issue #320, opened 2025-04-03
Backoff policy is constant, not exponential


9999years:
The current [backoff implementation](https://github.com/F1bonacc1/process-compose/blob/f87744d07985e7e941a272666c8ec3213f8b9e6b/src/app/process.go#L265-L271) in `process-compose` uses a (user-customizable) constant-time backoff with a minimum polling interval of 1 second.

However, users customizing a polling interval generally expect some form of [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) in order to poll more frequently at the beginning of a process (when changes are more likely) and then less frequently as time goes on.

Go seems to have a decent [backoff package by @cenkalti](https://pkg.go.dev/github.com/cenkalti/backoff/v4#BackOff) which exposes the following interface:

```go
type ExponentialBackOff struct {
	InitialInterval     time.Duration
	RandomizationFactor float64
	Multiplier          float64
	MaxInterval         time.Duration
	// After MaxElapsedTime the ExponentialBackOff returns Stop.
	// It never stops if MaxElapsedTime == 0.
	MaxElapsedTime time.Duration
	Stop           time.Duration
	Clock          Clock
	// contains filtered or unexported fields
}
```

PR #329, opened 2025-04-18
Add `process-compose project is-ready [--wait]`

9999years:
`process-compose --detached-with-tui --detach-on-success` is great for interactive uses, but not so great when scripting `process-compose`.

It's possible to wait for the project to be ready by parsing and polling the output of `process-compose process list --output=json`, but it's tedious and requires duplicating the success logic we already have in `process-compose`.

Therefore, this change adds a new command, `process-compose project is-ready`, which checks if all the processes are started and ready with the same logic as `--detach-on-success`. If the processes are not ready, `process-compose` prints an error and fails with exit code 1.

If `--wait` is given, `process-compose` polls the processes on a 1-second delay until they're ready:

    $ PC_LOG_FILE=/dev/stdout ~/process-compose/bin/process-compose project is-ready --wait
    25-04-18 16:12:44.026 INF Process Compose v1.46.0
    25-04-18 16:12:44.026 WRN Get "http://unix/processes": dial unix /tmp/mwb-MjA3wz5a/pc: connect: no such file or directory
    25-04-18 16:12:45.033 WRN Processes are not ready:
    • mailpit (health is Unknown)
    • postgres (health is Unknown)
    • temporal (health is Unknown)
    25-04-18 16:12:46.038 WRN Processes are not ready:
    • postgres (health is Unknown)
    25-04-18 16:12:47.041 INF 3 processes are ready

In the future, the delay (and a timeout) should be configurable; see issue #320.


src/cmd/is_ready.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/f1bonacc1/process-compose/src/client"
+	"github.com/f1bonacc1/process-compose/src/types"
+	"github.com/rs/zerolog/log"
+	"github.com/spf13/cobra"
+	"strings"
+	"time"
+)
+
+var (
+	waitReady = false
```
F1bonacc1, line 14: Can we move this to `src/config/Flags.go`?

F1bonacc1 (commented):
Thanks, @9999years,
That indeed can be very useful.
One small request and it is ready to merge.

src/cmd/is_ready.go
9999years: Done!

F1bonacc1 (approved):
Great, thanks!

PR #331, opened 2025-04-21
Add a `--dry-run` flag to validate the config

lopter:
Hello,

Thanks for the project, it is very useful.

One thing I have been missing is the ability to validate a generated config without starting any processes. This is very useful in CI, or with Nixpkgs when the config is generated as a derivation (and can be validated in the check phase of the derivation).

This PR attempts to add that, and I run into a couple questions I left in the code. Can you help answer them, and tell me if the overall approach is good?


F1bonacc1 (commented):
Hey @lopter,
Thanks for the PR! I think this is a very useful feature to have, and your approach is valid.

I think there is a simpler and more user-friendly way to tackle this issue.
If you add a `--dry-run` flag to the `up` and `root` commands, it will:
1. Consider additional (existing and future) flags when validating.
2. Eliminate the need for some of the refactoring.

What do you think?

lopter:
Thanks for the quick feedback @F1bonacc1, no strong feelings there, and I reworked the PR to go with your approach.

I was wondering if `--dry-run` flag should imply `is_struct: true`?

Is there anything we can do regarding validation of the values? The config file uses a fair amount of enumerations that would be nice to validate as well.

shell.nix
```diff
@@ -0,0 +1,11 @@
+{
```
F1bonacc1, line 1: Is this part of the PR or an accidental inclusion?

src/cmd/root.go
```diff
@@ -50,6 +50,11 @@ func runProjectCmd(args []string) {
 		_ = logFile.Close()
 	}()
 	runner := getProjectRunner(args, *pcFlags.NoDependencies, "", []string{})
+	if *pcFlags.DryRun {
+		processNames, _ := runner.GetLexicographicProcessNames()
+		fmt.Printf("Loaded %d processes from %d config files.\n", len(processNames), len(opts.FileNames)+len(opts.EnvFileNames))
```
F1bonacc1, line 55: Do you think that "Loaded" really reflects the intent of this feature?
Maybe something like "Successfully validated" will make more sense for the `--dry-run` flag?

F1bonacc1 (commented):
Thanks, @lopter, for the quick refactor.

Regarding values validation, I see 2 main approaches here:
1. Maintain `map[string]struct{}` maps for all the valid values for each Key. This is a quick and simple approach, but can easily go out of sync.
2. Add code generation that will be generating both the `const strings` and the validation maps from some `yaml` or `json` schema files. This sounds to me like not an easy task. But I had a short back and forth with Gemini about it, and it gave me a decent head start.

What do you think? Want to give it a try?

F1bonacc1:
> I was wondering if `--dry-run` flag should imply `is_struct: true`?
> 
Yes, I think this will be the right approach.

lopter:
Great, I can set `is_struct = true`.

Regarding validating values, I see your two suggestions, and I am wondering if a third option could be to implement enums. This way, values would be validated when the config files are unmarshalled?

src/cmd/root.go
lopter: Yes, I can do that.

shell.nix
lopter: Not accidental, but probably belong into another PR: if you don't mind I would like to add this file and wire it up from `flake.nix`. With a `shell.nix`, and maybe an `.envrc`, when a Nix user cd into the project, it sources everything needed to work on the project (that is even if they don't have golang installed on their machine).

lopter:
I see both `yaml.v3` and `yaml.v2` are used, is there a reason for that? Or could we move everything to `yaml.v3`?

lopter:
I did a quick test on one of the enums so that you can see what it would like, see: https://github.com/F1bonacc1/process-compose/pull/331/commits/6c74c74dd7906e2136674f2c394653baf539d3bd.

shell.nix
F1bonacc1: Sure, we can keep it.

F1bonacc1:
> I see both `yaml.v3` and `yaml.v2` are used, is there a reason for that? Or could we move everything to `yaml.v3`?

Yes, I think everything can be `yaml.v3`, thanks for noticing that.

F1bonacc1:
> Great, I can set `is_struct = true`.
> 
> Regarding validating values, I see your two suggestions, and I am wondering if a third option could be to implement enums. This way, values would be validated when the config files are unmarshalled?

I like your suggestion even more.
Let's go with that then.

lopter:
Ok, I see it has [`UnmarshalText`] but not sure how you can use it.

I did some rebase/fixup/squash operation, looks like I introduced some issue and one of the test gets stuck.

[`UnmarshalText`]: https://pkg.go.dev/github.com/rs/zerolog#Level.UnmarshalText

lopter:
Test is fixed: I had to move `RestartPolicyNo` to be the zero (`iota`) / default value.

lopter:
I updated the PR title, and had one last question: Should we also allow dry-run to be set through an environment variable?

F1bonacc1:
> I updated the PR title, and had one last question: Should we also allow dry-run to be set through an environment variable?

I don't think so, at least not until someone will ask for it 😃

lopter:
Alright, I don't have anything else then, this is ready to go for me :3

PR #339, opened 2025-05-23
table: color rows according to proc status

9999years:
At work, we use `--detach-on-success` (https://github.com/F1bonacc1/process-compose/pull/312) to automatically exit the `process-compose` TUI when processes finish starting up. When processes fail to start up, engineers often don't notice that a process has failed, and instead assume that `process-compose` has a bug.

The UI makes it pretty easy to miss a failed process; the icon on the left of the table is red, but that's about it. The exit code is off to the right, and it's not immediately obvious that a 1 means there's a problem.

We can make the UI significantly more legible by coloring rows according to the process state. We'll reuse the color of the state icon for this. This displays nicely when rows are selected as well; a selected failed process will show up with black text on a red background, making the failure very clear (as opposed to the previous default blue background).

### Before

image

### After

image

### DIFF
--- a/src/tui/proc-table.go
+++ b/src/tui/proc-table.go
@@ -403,9 +403,9 @@ func (pv *pcView) getIconForState(state types.ProcessState) (string, tcell.Color
 		if state.ExitCode == 0 {
 			return "●", pv.styles.ProcTable().FgCompleted.Color()
 		}
-		return "❌", pv.styles.ProcTable().FgError.Color()
+		return "✘", pv.styles.ProcTable().FgError.Color()
 	case types.ProcessStateError:
-		return "❌", pv.styles.ProcTable().FgError.Color()
+		return "✘", pv.styles.ProcTable().FgError.Color()
 	case types.ProcessStateDisabled,
 		types.ProcessStateForeground:
 		return "◯", pv.styles.ProcTable().FgPending.Color()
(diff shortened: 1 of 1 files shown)


9999years:
@F1bonacc1 I think the icons are fun, honestly! I wouldn't be heartbroken if they were removed. But if I were picking the icons myself they would probably _all_ be emoji, so you may wish to take my opinions with a grain of salt! 🧂😉

F1bonacc1:
> @F1bonacc1 I think the icons are fun, honestly! I wouldn't be heartbroken if they were removed. But if I were picking the icons myself they would probably _all_ be emoji, so you may wish to take my opinions with a grain of salt! 🧂😉

Hmm... Now it's an even more difficult decision.
Can we agree that they should be either all emojis or all ASCII?

Otherwise, it drives my OCD crazy 😄 

image

9999years:
I think that's probably because — and I'm embarrassed to admit this — my terminal is using the non-monospaced variant of [PragmataPro](https://fsd.it/shop/fonts/pragmatapro/), where some symbols are wider than others... let me see about this, I don't _think_ any of the codepoints used have emoji presentation.

9999years:
OK, according to [`emoji-data.txt`](https://unicode.org/Public/UNIDATA/emoji/emoji-data.txt), ❌ U+274C CROSS MARK _is_ an emoji. I think the rest are not though.

9999years:
OK, I replaced ❌ U+274C CROSS MARK with ✘ U+2718 HEAVY BALLOT X so that none of the status symbols have emoji presentation. Here's what it looks like now, with PragmataPro Mono:

image

I recently switched terminal emulators from iTerm2 to WezTerm, so the color scheme is a bit different this time, forgive me for this :)

PR #332, opened 2025-04-22
feat: add script to generate json schema for yamlls to allow for autocompletion

harryvince:
Adds support for autocompletion by [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).

Added a new schema command under the docs section that will auto generate a json schema that can be used to validate keys within the yaml.

Users can then either add the schema validation to their file by doing the following:

Adding this comment at the top of their file: 
```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/F1bonacc1/process-compose/main/schema.json
```

or by adding the schema url to their respective config [vscode](https://www.ibm.com/docs/en/adffz/dbb/3.0.0?topic=ide-configuring-schema-validation-vscode) or [neovim](https://github.com/neovim/nvim-lspconfig/blob/5af60bbb835d5b6efdc64cdff2a5b27cde4120de/doc/configs.md#yamlls)

PR #356, opened 2025-06-25
fix: add process struct tags to help yamlls recognize non-required fields

adasauce:
addresses comments in #332 about misrepresenting required fields.


F1bonacc1 (approved):
Looks good. 
Ready to merge once the conflicts are resolved.

adasauce:
rebased on main and re-ran make schema to resolve the conflict.

note: it might help for maintainability if the json schema were pretty printed so it would be potentially easier to identify sources of change even though this is effectively a build artifact.

F1bonacc1:
> note: it might help for maintainability if the json schema were pretty printed so it would be potentially easier to identify sources of change even though this is effectively a build artifact.

Totally agree with that.
I think if we change this to `MarshalIndent` it will fix that:
https://github.com/F1bonacc1/process-compose/blob/2cc46d1e72c2418fd34ea9e76fd1802c89fbaa6d/src/cmd/docs.go#L43

adasauce:
happy to have a look at it in another PR to keep it targeted.

F1bonacc1:
> happy to have a look at it in another PR to keep it targeted.

Makes sense. Thank you!

PR #359, opened 2025-07-02
fix: reverse shutdown order does not contain all dependencies

fabiankranewitter-aqt:
This bug affects the shutdown order when multiple dependencies and the option  `--ordered-shutdow ` are used. This change fixes the `reverseDependencies` list to include all dependencies and the shutdown order then works smoothly with multiple dependencies.


F1bonacc1:
Hi @fabiankranewitter-aqt,
The test that you added fails. Can you please check before I review the PR.

F1bonacc1:
Looks good, thank you!

PR #363, opened 2025-07-04
Fix: if a process will never be ready, its dependants should fail

maddiehorger-fr:
You can see this in more detail with the fixture I provided, but if a process has a `ready_log_line` which can never be reached (process is already completed), then its dependents should not get stuck waiting for it to be ready.

I don't know if this patch really adequately covers all the use cases it ought to cover, but it covers the one in the fixture, which is representative of my use case. If I _have_ missed any related scenarios, let's figure out what we need to do to cover them, tests included!


F1bonacc1:
Great!
Good catch, thank you!

PR #369, opened 2025-07-20
feat: Allow configuring log level through env

### LINKED ISSUE #366: Config path warning on every startup
I'm getting the following warning on every startup:

```
{"level":"warn","error":"could not locate `process-compose` in any of the following paths: [~/.config /etc/xdg ~/.nix-profile/etc/xdg /nix/profile/etc/xdg ~/.local/state/nix/profile/etc/xdg /etc/profiles/per-user/happens/etc/xdg /nix/var/nix/profiles/default/etc/xdg /run/curr
ent-system/sw/etc/xdg]","time":"2025-07-09T12:36:36+02:00","message":"Path not found for process compose config home"}
{"level":"warn","error":"could not locate `process-compose` in any of the following paths: [~/.config /etc/xdg ~/.nix-profile/etc/xdg /nix/profile/etc/xdg ~/.local/state/nix/profile/etc/xdg /etc/profiles/per-user/happens/etc/xdg /nix/var/nix/profiles/default/etc/xdg /run/curr
ent-system/sw/etc/xdg]","time":"2025-07-09T12:36:36+02:00","message":"Path not found for process compose config home"}
```

From what I can see, global config is not actually loaded and process-compose runs fine without it. I think this could be a `debug` log instead, since a warning implies that an action should be taken.

happenslol:
Closes #366.

This allows setting the log level through `PC_LOG_LEVEL`. I chose to abort if an invalid log level is passed, since I wouldn't expect this to occur unless the user specifically set the env var and would want to be notified if it's invalid. We could alternatively emit an error log here if you want.

Additionally, the warning log when no global config is found is changed to a debug log instead.


F1bonacc1 (commented):
Thank you @happenslol,
The code looks good and indeed addresses the issue.
But there are 2 potential issues:
1. In case there is project level log level, the `PC_LOG_LEVEL` variable should override that.
2. I think this won't happen, because the location in which you parse the env variable is not in the right place. I think it should be here:
https://github.com/F1bonacc1/process-compose/blob/55d64379d574e06558630657918878ff8952acb9/src/loader/validators.go#L24-L38

This will also allow you to reuse some of the things that already exit there.

happenslol:
That makes it a lot easier! Is this better, or should I put the code inside a mutator function that is then applied?

src/loader/loader.go
```diff
@@ -51,6 +52,11 @@ func Load(opts *LoaderOptions) (*types.Project, error) {
 	// If DryRun is set to validate the config, then force IsStrict to true:
 	mergedProject.IsStrict = opts.DryRun || mergedProject.IsStrict
 
+	// Override log level if given in env
+	if envLevel, set := os.LookupEnv(config.LogLevelEnvVarName); set {
+		mergedProject.LogLevel = envLevel
```
F1bonacc1, line 57: The level should be validated, please do it at the validators.go location that I sent you.

F1bonacc1: My bad.
You are right.
Note to self: don't check PRs while doing 7 other things 😄

happenslol:
@F1bonacc1 Any chance you could cut a release soon, including this?

Issue #343, opened 2025-05-29
feat(rust): allow rust to launch process-compose with all parameters


dzmitry-lahoda:
## Feature Request

Currently Rust can create project files and work with running process-compose via https://github.com/F1bonacc1/process-compose/tree/main/crates .

So remaining integration point is to start new process-compose via `command line`,

specifically via `process-wrap` crate.

#### Use Case:

Rust has not good slim xplat process manager.
So with good Rust support in process-compose, it may thing it have.

`process-compose` is static binary and Rust can (optionally) download files as it does in other well known popular crates. 

So just adding crate of process-compose to project, can make it full blown near native process manager for Rust developers.

#### Proposed Change:

Add Rust builder crate to allow instance new process-compose via `process-wrap` with all relevant and possible startup flags (as in go).

#### Who Benefits From The Change(s)?

Rust developers.

#### Alternative Approaches

Nobody seems bothered to write in pure Rust ;)

dzmitry-lahoda:
I feel this can be PRed by AI (or at least generated and maintained in sync with Go). Will try this.

dzmitry-lahoda:
will donate 50USDC in crypto for required part implemented

PR #374, opened 2025-08-05
feat(rust): strongly typed cli interface to start process compose

dzmitry-lahoda:
- for #343

- up to user (or other PR) make it nicer daemon like interface (no tui, port is required, etc)


F1bonacc1:
Thank you @dzmitry-lahoda!
Also thank you, @Javyre and @beborngod for reviewing.
I am not familiar with rust, and I always prefer an additional pair of eyes going through the code before merging.

PR #380, opened 2025-08-24
app, config: Obey parentOnly for Windows and add same log as for Unix and fix forceKillOnTimeout

vandot:
- Obey parentOnly for Windows and add same log as for Unix
- In case of shutdown timeout, per [docs](https://github.com/F1bonacc1/process-compose/blob/main/www/docs/launcher.md?plain=1#L176), the process group should receive the `SIGKILL` signal (irrespective of the `shutdown.parent_only` option)

### DIFF
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -436,7 +436,7 @@ func (p *Process) forceKillOnTimeout() error {
 		return nil
 	case errors.Is(err, context.DeadlineExceeded):
 		log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
-		return p.command.Stop(int(syscall.SIGKILL), p.procConf.ShutDownParams.ParentOnly)
+		return p.command.Stop(int(syscall.SIGKILL), false)
 	default:
 		log.Error().Err(err).Msgf("terminating %s with timeout %d failed", p.getName(), p.procConf.ShutDownParams.ShutDownTimeout)
 		return err
--- a/src/command/stopper_windows.go
+++ b/src/command/stopper_windows.go
@@ -14,10 +14,10 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
 		Msg("Stop Windows process.")
 
 	if parentOnly {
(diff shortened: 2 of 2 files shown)


src/app/process.go
```diff
@@ -436,7 +436,7 @@ func (p *Process) forceKillOnTimeout() error {
 		return nil
 	case errors.Is(err, context.DeadlineExceeded):
 		log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
-		return p.command.Stop(int(syscall.SIGKILL), p.procConf.ShutDownParams.ParentOnly)
+		return p.command.Stop(int(syscall.SIGKILL), false)
```
F1bonacc1, line 439: Why not to respect the process config on timeout?

src/command/stopper_windows.go
```diff
@@ -5,9 +5,19 @@ import (
...
 	//p.command.Process.Kill()
-	kill := exec.Command("TASKKILL", "/T", "/F", "/PID", strconv.Itoa(c.Pid()))
+	log.
+		Debug().
+		Int("pid", c.Pid()).
+		Bool("parentOnly", parentOnly).
+		Msg("Stop Windows process.")
+
+	if parentOnly {
+		kill := exec.Command("C:\Windows\System32\taskkill.exe", "/F", "/PID", strconv.Itoa(c.Pid()))
```
F1bonacc1, line 17: Why do we need an absolute path here?

src/command/stopper_windows.go
```diff
@@ -5,9 +5,19 @@ import (
...
+		return kill.Run()
+	}
+	kill := exec.Command("C:\Windows\System32\taskkill.exe", "/T", "/F", "/PID", strconv.Itoa(c.Pid()))
```
F1bonacc1, line 20: Why do we need an absolute path here?

src/command/stopper_windows.go
vandot: SonarQube complained about a relative path https://github.com/F1bonacc1/process-compose/runs/48774929711
but unfortunately after issue is fixed previous report is no longer accessible.
I can put it back so you check the report?

PR #385, opened 2025-08-25
feat(api): namespace wide disable/enable/stop operations on processes

dzmitry-lahoda:


dzmitry-lahoda:
As for testing, I will do Rust CLI and write Rust tests in our project. If any issues will return with PR.

dzmitry-lahoda:
@F1bonacc1 I do not think it can be merged, so some design could be discussed?

Why something like this to be in Process Compose.
- to use PC remotely, behind some automation, in tests (or other transient jobs), seems need something to be. so it is not one user use case, but agenda for many. i provided design and impl of one of option.
- making own things on client side, can end up build own mutex covering whole PC OpenAPI methid, that is harder than should be and harder to maintain. providing what API can as part of Go and CLI clients later will need even more wrapping boiler plate. so making outside is hard.
- error handling. i feel that error handling to be improved (made more typed overtime). changes can only start in PC code, not string contains by outside code.

src/api/pc_api.go
```diff
@@ -208,6 +208,210 @@ func (api *PcApi) StopProcesses(c *gin.Context) {
...
+// @Description	Merge processes from a partial config; all must share the same namespace/
+// @Tags			Namespace
+// @Summary		Post config fragment with processes (single namespace)
+// @Accept		json
+// @Produce		json
+// @Param			processes	body	types.Processes	true	"One or more processes, all in the same namespace"
+// @Success		200		{object}	map[string]string	"All updated"
+// @Failure		400		{object}	map[string]string  "Some processes failed to be updated. Returns error if all failed, else returns success and failures map"
+// @Router			/namespace [post]
+func (api *PcApi) UpdateNamespace(c *gin.Context) {
```
F1bonacc1, line 364: This looks redundant - the namespace is part of the process - this should handle **all** the requested processes. It's on the client to make sure that all the processes belong to the same NS.

src/api/pc_api.go
```diff
@@ -208,6 +208,210 @@ func (api *PcApi) StopProcesses(c *gin.Context) {
...
+// @Description	Enables all processes in the given namespace
+// @Tags			Namespace
+// @Summary		Enable all processes in a namespace
+// @Produce		json
+// @Param			name	path		string				true	"Namespace Name"
+// @Success		200		{object}	map[string]string	"All processes in namespace enabled"
+// @Success		400		{object}	map[string]string	"Some processes in namespace failed to be enabled"
+// @Failure		404		{object}	map[string]string	"No processes in namespace"
+// @Router			/namespace/enable/{name} [patch]
+func (api *PcApi) EnableNamespace(c *gin.Context) {
```
F1bonacc1, line 311: Too much logic in the API, should be moved to the project runner.

src/api/pc_api.go
```diff
@@ -208,6 +208,210 @@ func (api *PcApi) StopProcesses(c *gin.Context) {
...
+// @Description	Disables all processes in the given namespace
+// @Tags			Namespace
+// @Summary		Disable all processes in a namespace
+// @Produce		json
+// @Param			name	path		string				true	"Namespace Name"
+// @Success		200		{object}	map[string]string	"All processes in namespace disabled"
+// @Success		400		{object}	map[string]string	"Some processes in namespace failed to be disabled, can happen if several opposite updates to same namespace are happening"
+// @Failure		404		{object}	map[string]string "No processes in namespace"
+// @Router			/namespace/disable/{name} [patch]
+func (api *PcApi) DisableNamespace(c *gin.Context) {
```
F1bonacc1, line 258: Too much logic in the API, should be moved to the project runner.

src/api/pc_api.go
```diff
@@ -208,6 +208,210 @@ func (api *PcApi) StopProcesses(c *gin.Context) {
...
+// @Tags			Namespace
+// @Summary		Stop all processes in a namespace
+// @Produce		json
+// @Param			name	path		string				true	"Namespace Name"
+// @Success		200		{object}	map[string]string	"Stopped All Processes in Namespace"
+// @Success		207		{object}	map[string]string	"Stopped Part of Processes in Namespace"
+// @Failure		400		{object}	map[string]string   "Failed to stop some processes, they may have some dependants"
+// @Failure		404		{object}	map[string]string   "No proccesses in namespace"
+// @Router			/namespace/stop/{name} [patch]
+func (api *PcApi) StopNamespace(c *gin.Context) {
```
F1bonacc1, line 223: Too much logic in the API, should be moved to the project runner.

F1bonacc1 (commented):
I think this is a good idea.
This make NSs a more important concept in the PC than it is today.
Thanks for the initiative.

I left you a few comments about the implementation.
Most of them are related to the business logic being implemented in the API instead of the `project runnner`.

Can you please also add some tests? I am not sure that there won't be any race issues. Check with `make testrace`.

src/api/pc_api.go
dzmitry-lahoda: Ok. Sound good.

dzmitry-lahoda:
> This make NSs a more important concept in the PC than it is today.

Yeah, people can add more things too later. For example TUI grouping or depends_on namespace or like.

> Can you please also add some tests? I am not sure that there won't be any race issues. Check with make testrace.

Ok will do tests.

dzmitry-lahoda:
@F1bonacc1 are tests for such changes usually written on api level or on project runner level?

dzmitry-lahoda:
As of now "testing" on API(OpenAPI) level via Rust in our test runner codebase. Will send some patches if any after.

F1bonacc1:
> @F1bonacc1 are tests for such changes usually written on api level or on project runner level?

Project runner level. See `system_test.go`. We want to check concurrency issues (`make testrace`).

dzmitry-lahoda:
need to add namespace ready after https://github.com/F1bonacc1/process-compose/pull/396, so that I can make test(test = namespace) setup wait ready

dzmitry-lahoda:
`ready` seems hard to define. so just need to return 404 when there is no such namespace at least for now.

impl of not ready or other can be decided by client (handling status code + state + config).

Issue #386, opened 2025-08-25
chore(api): add schemas of return JSON objects for 200 OK code


dzmitry-lahoda:
so that Rust can generate proper types for immediate use

PR #387, opened 2025-08-25
chore(api): better OpenAPI returns and WS route declaration

dzmitry-lahoda:
for #386 and #379


dzmitry-lahoda:
@F1bonacc1 please take a look. Need to use PC from typed languages.

F1bonacc1 (approved):
LGTM
Thanks

PR #390, opened 2025-08-28
feat: use envsubst for more feature filled environment expansion than os.Expand

adasauce:
allows for more flexible process-compose files that get baked into containers and such.

i.e.

```
version: "0.5"
log_level: "trace"
vars:
    VIRTUAL_ENV: "${PC_VIRTUAL_ENV:-./.venv}"
 ```

so some sane defaults can be provided or overridden by environment vars.
I've also used it for pattern replacement with like `--allowed-hosts=${HOST%:8000}`

been using this feature in my own deployment for the last month or so to success and thought you might like it as well.


adasauce:
ah I see there's a make command that is supposed to update some nix hashes, but I don't have any nix related tooling on my machine to run it.

dzmitry-lahoda:
@adasauce you can download, chmox +x and run https://manpages.ubuntu.com/manpages/jammy/man1/nix-hash.1.html (all binaries are here https://launchpad.net/ubuntu/jammy/+package/nix-bin)

and run `./scripts/update-vendor-hash.sh` from this repo

so you not need full nix nor really need installs.

adasauce:
> @adasauce you can download, chmox +x and run https://manpages.ubuntu.com/manpages/jammy/man1/nix-hash.1.html (all binaries are here https://launchpad.net/ubuntu/jammy/+package/nix-bin)
> 
> and run `./scripts/update-vendor-hash.sh` from this repo
> 
> so you not need full nix nor really need installs.

thanks for the tip. I ended up just piping it in through the nix docker image 😶‍🌫️

F1bonacc1 (commented):
Nice! Thanks a lot!
I didn't know about this package.

Would you mind adding some documentation with a reference to the package README?

PR #393, opened 2025-08-29
chore(rust): fixed WS support in Rust and cleanup defaults in toml

dzmitry-lahoda:


dzmitry-lahoda:
@F1bonacc1 please take a look, some fixes from our internal usage.

F1bonacc1 (approved):

PR #401, opened 2025-09-09
chore(rust): builder support for openapi

dzmitry-lahoda:


dzmitry-lahoda:
works well for us. @F1bonacc1 may you check?

it just adds proper feature to add constructor-builder API to generated code

dzmitry-lahoda:
published https://crates.io/crates/process-compose-client, next version seems will have to diverge from upstream to get features

Issue #317, opened 2025-03-27
Environment variables from .env should not be passed to child processes by default


sciyoshi:
## Feature Request

I think the default behavior of process-compose should be to use the variables from .env for interpolation within the configuration file, but that those should not be passed to child processes by default.

The reason for this is that many of the services that are run as child processes have their own logic and configuration for loading of .env files. The problem with process-compose providing them is that the environment normally takes precedence, and means that making changes to the .env file requires a whole process-compose restart for it to be picked up (rather than just restarting the child process).

I realize this would be a breaking change, but I think this is a better default behavior. The current behavior could be available as a flag or as a configuration on a per-service basis.

#### Proposed Change:

By default, .env will be read and parsed by process-compose, but will not update the actual environment variables.

F1bonacc1:
Hey @sciyoshi,

Now that you mention it, I see how .env loading can create an undesired behavior for processes that might have their own, same env variables (e.g. PASSWORD).
Before I go ahead and implement only config interpolation with `.env` without actually loading it into environment variables, I want to be sure that you considered using Process Compose [Vars](https://f1bonacc1.github.io/process-compose/configuration/#variables) and [Environment](https://f1bonacc1.github.io/process-compose/configuration/#environment-variables) instead.
Those will allow you to:
1. Local (to `process-compose.yaml`) only interpolation (with Vars).
2. Reloading environment variables without restarting Process Compose (`Ctrl+L`).
3. Env variables isolation between processes (with `processes.process.environment`).

sciyoshi:
Thanks @F1bonacc1! Yes, I had considered these - however our use case is env variables that need to be accessed by multiple services and we're using .env files to specify them. Reloading with `Ctrl+L` seems to read new env variables that are added to the .env but does not overwrite existing ones, probably for the same reason as what this PR is saying (that the one already set in the environment takes precedence over the one from the .env file on reload).

F1bonacc1:
Hey @sciyoshi,

Please help me to understand your scenario a bit better. I am not sure how those 2 statements can be true at the same time:

> I think the default behavior of process-compose should be to use the variables from .env for interpolation within the configuration file, but that those should not be passed to child processes by default.

> our use case is env variables that need to be accessed by multiple services and we're using .env files to specify them

sciyoshi:
[...]

PR #417, opened 2025-11-13
fix: restore environment variable precedence and align with docker-compose

### LINKED ISSUE #416: [Regression] Environment variable precedence broken in v1.63.0 - .env overrides YAML
## Pre-Submission Checklist

- [x] Relevant configuration snippet (see below)
- [x] Process-compose logs from `process-compose info`
- [x] Minimal, Complete, and Verifiable example (see reproduction steps)

## Environment

**Process-Compose Version:**
```
Version:        v1.76.1
Commit:         dc18cb4
Date (UTC):     2025-11-05T22:11:21Z
```

**Operating System:** Linux (also tested on Linux 6.12.50-1-lts)
**Architecture:** x86_64

**Process-Compose Info:**
```
Process Compose
Logs:           /tmp/process-compose-manolo.log
Custom Theme:   /home/manolo/.config/process-compose/theme.yaml
Settings:       /home/manolo/.config/process-compose/settings.yaml
```

## Description

Environment variables defined in YAML configuration sections (`environment:`) are being overridden by `.env` file variables, which contradicts the documented behavior and represents a **regression** introduced in **v1.63.0**.

According to the official documentation and expected behavior, YAML `environment:` sections should have **higher precedence** than `.env` files. However, the current implementation does the opposite.

### 1. Create a minimal test case

**File: `.env`**
```bash
TEST_VAR=from_dotenv
```

**File: `process-compose.yaml`**
```yaml
version: "0.5"

# Global environment section should override .env
environment:
  - TEST_VAR=from_global_yaml

processes:
  # Test 1: .env vs global environment
  test-global-override:
    command: "bash -c 'echo \"Global test - TEST_VAR=$TEST_VAR\"'"

  # Test 2: .env vs local environment
  test-local-override:
    command: "bash -c 'echo \"Local test - TEST_VAR=$TEST_VAR\"'"
    environment:
      - TEST_VAR=from_local_yaml
```

### 2. Run process-compose

```bash
PC_DISABLE_TUI=1 process-compose up
```

## Expected Behavior

According to the documented precedence order, YAML `environment:` sections should override `.env` file values:

```
Precedence (highest to lowest):
1. Process-level environment: section
2. Global environment: section
3. .env file
4. System environment
```

**Expected output:**
```
[test-global-override] Global test - TEST_VAR=from_global_yaml
[test-local-override ] Local test - TEST_VAR=from_local_yaml
```

## Actual Behavior

The `.env` file overrides both global and local YAML `environment:` sections:

**Actual output:**
```
[test-global-override] Global test - TEST_VAR=from_dotenv
[test-local-override ] Local test - TEST_VAR=from_dotenv
```

## Root Cause Analysis

After analyzing the source code and testing multiple versions, I've identified this as a **regression introduced in v1.63.0** (May 5, 2024) by PR #317.

### Version Testing Results

I tested 6 versions spanning 8 months:

| Version | Release Date | Behavior | Status |
|---------|--------------|----------|--------|
| v1.40.0 | 2023-11-09 | YAML wins over .env | ✅ Correct |
| v1.46.0 | 2023-12-20 | YAML wins over .env | ✅ Correct (last known good) |
| **v1.63.0** | **2024-05-05** | **.env wins over YAML** | **❌ Regression introduced** |
| v1.64.1 | 2024-05-10 | .env wins over YAML | ❌ Bug persists |
| v1.76.1 | 2024-11-05 | .env wins over YAML | ❌ Bug persists (current) |

### Technical Details

In PR #317, the `.env` loading mechanism was refactored from `godotenv.Load()` (which modifies `os.Environ()`) to `godotenv.Read()` (which returns a map).

The new implementation appends `.env` variables **last** in the environment building chain:

```go
// Approximate code from src/process/process.go
env = append(env, os.Environ()...)           // 1. System environment
env = append(env, p.globalEnv...)            // 2. Global YAML
env = append(env, p.procConf.Environment...) // 3. Process YAML

// .env appended LAST → gives it highest precedence
if p.dotEnvVars != nil && !p.procConf.DisableDotEnv {
    for k, v := range p.dotEnvVars {
        env = append(env, k+"="+v)
    }
}
```

Since the last `append()` wins in Go when there are duplicate keys, `.env` variables now have the **highest precedence**, which is the opposite of the intended behavior.

## Impact

This regression breaks several common use cases:

### 1. Cannot override .env defaults with YAML
```yaml
# .env
DATABASE_HOST=production-db

# process-compose.yaml
environment:
  - DATABASE_HOST=localhost  # ← IGNORED due to bug
```

### 2. Process-specific configuration doesn't work
```yaml
processes:
  service-1:
    environment:
      - PORT=3000  # ← IGNORED if PORT exists in .env
```

### 3. Docker Compose migrations break
Docker Compose correctly implements YAML > .env precedence, so configurations migrated to process-compose exhibit unexpected behavior.

## Workarounds

Until this is fixed, users can work around the issue by:

1. **Disable .env globally:** `process-compose --disable-dotenv`
2. **Disable .env per-process:** Use `is_dotenv_disabled: true`
3. **Avoid .env for configurable variables:** Only use `.env` for immutable constants
4. **Downgrade to v1.46.0:** Last version with correct behavior

## Proposed Fix

The fix should restore the correct precedence order by appending `.env` variables **before** YAML environment sections:

```go
// System and .env first (lower precedence)
env = append(env, os.Environ()...)
if p.dotEnvVars != nil && !p.procConf.DisableDotEnv {
    for k, v := range p.dotEnvVars {
        env = append(env, k+"="+v)
    }
}

// YAML after (higher precedence)
env = append(env, p.globalEnv...)
env = append(env, p.procConf.Environment...)
```

## Additional Context

- This issue affects all users who expect to override `.env` defaults with YAML configuration
- The bug has persisted for 6+ months (v1.63.0 to v1.76.1)
- PR #317 successfully added the `is_dotenv_disabled` feature, but inadvertently broke precedence
- Documentation still reflects the correct (but not implemented) precedence order

## Testing Recommendation

I recommend adding regression tests to prevent this in the future:

```go
func TestEnvironmentPrecedence(t *testing.T) {
    // Given: .env has TEST_VAR=from_dotenv
    // And: YAML global has TEST_VAR=from_global
    // When: Process starts
    // Then: TEST_VAR should equal "from_global" (YAML wins)
}
```

## References

- **Introducing commit:** f9010f7 (v1.63.0)
- **Related PR:** #317 "Added ability to disable passing .env vars to processes"
- **Related issue:** #317 "Environment variables from .env should not be passed to child processes by default"

---

I'm happy to provide additional information, test cases, or submit a PR with the fix if that would be helpful.

ManuelLR:
Fixes two environment variable precedence issues to restore correct behavior and align with Docker Compose standards.

### Issues Fixed

1. **Regression from v1.63.0** - .env file variables incorrectly override YAML environment sections
2. **Docker Compose alignment** - System/shell environment variables now override .env file values

#### Problem 1: .env overriding YAML (Regression)
In v1.63.0 (PR #317), the .env loading refactor inadvertently changed the precedence order. The .env variables were appended last, giving them highest precedence instead of lowest.

**Expected:** System → .env → Global YAML → Local YAML  
**Broken:** System → Global YAML → Local YAML → .env

#### Problem 2: .env overriding shell variables
Previously, .env file variables had higher precedence than shell environment variables, preventing runtime overrides.

**Example:**
```bash
# .env contains: PORT=3000
PORT=8080 process-compose up  # Should use 8080, was using 3000
```

### The Fix

Reorders environment variable appends in `getProcessEnvironment()`:

**Correct precedence (lowest → highest):**
1. .env file variables (baseline defaults)
2. System environment (shell overrides - matches docker-compose)
3. Global YAML environment section
4. Local process YAML environment section (highest)

This matches Docker Compose behavior where shell variables override .env files.

### Testing

- Added 13 comprehensive regression tests covering all precedence scenarios
- All 260 tests pass with race detection
- Linting passes (0 issues)
- Tests use `t.Setenv()` following Go best practices
- Preserves `is_dotenv_disabled` feature from PR #317

Closes #416

### DIFF
--- a/src/app/process_env_test.go
+++ b/src/app/process_env_test.go
@@ -9,8 +9,8 @@ import (
 
 // Test_getProcessEnvironment_Precedence tests the environment variable precedence order.
 // The correct order (lowest to highest) should be:
-// 1. System environment (os.Environ)
-// 2. .env file variables
+// 1. .env file variables
+// 2. System environment (os.Environ)
 // 3. Global YAML environment section
 // 4. Local process YAML environment section (highest)
 func Test_getProcessEnvironment_Precedence(t *testing.T) {


ManuelLR:
Hey @F1bonacc1!

Thanks so much for reviewing and approving this! Really appreciate the quick turnaround.

I wanted to mention one thing I noticed while comparing with Docker Compose's behavior - not sure if it's intentional or worth addressing:

## Shell Environment vs .env Precedence

I was looking at [Docker Compose's environment variable precedence docs](https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/) and noticed a difference in how shell variables vs .env files are handled.

**Example scenario:**

.env file contains:
```bash
PORT=3000
```

User runs with shell override:
```bash
PORT=8080 process-compose up
```

Current behavior: Uses 3000 (from .env)  
Docker Compose behavior: Would use 8080 (shell wins)

---

In Docker Compose, shell/system environment variables have higher precedence than the .env file, but in process-compose it's the opposite (since .env is appended after `os.Environ()`).

Since the README mentions process-compose "tries to follow the docker-compose specifications," I wanted to check if this difference is intentional or if it would make sense to align it?

**If you'd like me to fix this, it's a simple change** - I can swap the order in `getProcessEnvironment()` so .env is appended before `os.Environ()` instead of after. Happy to push that update right now so everything is fixed in this PR at once.

Let me know what you prefer!

Thanks again for the great project!

F1bonacc1:
I think that's a very good suggestion.
This approach follows closer the industry standard tool and also allows more flexibility to override env values without editing files.
Let's put everything in the same PR.
Thank you!

F1bonacc1 (approved):
Looks great!
There is only one small suggestion to one of the comments, to be better aligned with the precedence order.

PR #435, opened 2026-01-19
fix: go.mod parses to 1_24_0 instead of nixpkg's 1_24

jessedhillon:
This overlay allows our `swag2op` Go module to map to Go 1.24.0 (whose nixpkg entry is `go_1_24` instead of `go_1_24_0` which we generate)


F1bonacc1 (approved):
LGTM
TYVM!

PR #438, opened 2026-02-12
feat: add JSON pretty-print toggle for TUI logs

EricWardner:
## Summary                                                                                                                                                   
                                                          
  - Add a runtime toggle (`p` key) to pretty-print JSON log lines in the TUI                                                                                   
  - When enabled, valid JSON lines are formatted with newlines and 2-space indentation
  - Toggling re-renders existing buffered logs so the effect is immediate                                                                                      
  - Off by default, global toggle matching the pattern of other runtime toggles (wrap, follow)                                                                 

  ## Details

  ### Insertion point

  The pretty-print transformation is applied at the top of `LogView.WriteString()`, before the existing ANSI/error-coloring pipeline. This means it works correctly with both ANSI and non-ANSI code paths — JSON is expanded first, then the normal formatting (escaping, error highlighting) applies to each resulting line.

### JSON detection approach

  A fast heuristic check (`trimmed[0] == '{' || trimmed[0] == '['`) gates the `json.Unmarshal` call so non-JSON lines skip the allocation entirely. `json.RawMessage` is used instead of `interface{}` to avoid unnecessary reflection — we just need to re-indent, not inspect the structure.

  Considered avoiding `json.Unmarshal` altogether (e.g. regex-based detection or only checking for a leading `{`), but anything short of a real parse would either miss valid JSON or false-positive on non-JSON lines. Since this only runs when the user explicitly toggles the feature on, the cost is acceptable — it's a deliberate opt-in, and the heuristic prefix check still short-circuits the common case of plain text lines.

  ### JSON detection approach

  A fast heuristic check (`trimmed[0] == '{' || trimmed[0] == '['`) gates the `json.Unmarshal` call so non-JSON lines skip the allocation entirely. `json.RawMessage` is used instead of `interface{}` to avoid unnecessary reflection — we just need to re-indent, not inspect the structure.

  ### Re-rendering on toggle

  When the user presses `p`, the handler calls `unFollowLog()` + `followLog()` which clears the view and re-subscribes to the process log buffer. This causes all buffered lines to flow back through `WriteString()` with the new setting, so existing logs are re-rendered immediately rather than only affecting new lines.

  ### Keybinding choice

  `p` was chosen as a mnemonic rune shortcut (pretty-print), following the existing pattern of single-character shortcuts (`m` for mark, `n` for namespace). The toggle appears in the help footer bar between "Wrap" and "Select" with "Pretty On"/"Pretty Off" labels.

### DIFF
--- a/src/tui/view.go
+++ b/src/tui/view.go
@@ -534,7 +534,6 @@ func (pv *pcView) updateHelpTextView() {
 	pv.shortcuts.addToggleButton(ActionLogScreen, pv.helpFooter, logScrBool)
 	pv.shortcuts.addToggleButton(ActionFollowLog, pv.helpFooter, !pv.logFollow)
 	pv.shortcuts.addToggleButton(ActionWrapLog, pv.helpFooter, !pv.logsText.IsWrapOn())
-	pv.shortcuts.addToggleButton(ActionLogPrettyPrint, pv.helpFooter, !pv.logsText.IsPrettyPrintOn())
 	if config.IsLogSelectionOn() {
 		pv.shortcuts.addToggleButton(ActionLogSelection, pv.helpFooter, !pv.logSelect)
 	}


EricWardner:
Great! I think I removed the right line. Thanks for the review!

PR #450, opened 2026-03-26
fix: reap zombie processes

54rt1n:
Fixes an issue where zombie process handles don't terminate cleanly and get stuck, unable to restart.

### DIFF
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -187,6 +187,14 @@ func (p *Process) waitForStdOutErr() {
 	ctx, cancel := context.WithCancel(context.Background())
 	if p.procConf.IsDaemon {
 		ctx, cancel = context.WithTimeout(context.Background(), time.Duration(p.procConf.LaunchTimeout)*time.Second)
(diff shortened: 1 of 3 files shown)


F1bonacc1:
Hey @54rt1n,

Thank you for the PR, it looks valid and very surgical.
Do you know if there is a way to replicate the zombie / terminating state you are describing? I want to try to replicate it in my environment.

54rt1n:
This is something that happens when a parent process exits, but a child process still has an output pipe open (stdout/stderr). Since the child never allows the EOF to propagate, the Wait is never hit so the process isn't reaped and is left in Terminating.

F1bonacc1:
Thanks again for finding such a complex issue @54rt1n!

I refactored it a little bit here https://github.com/F1bonacc1/process-compose/commit/18f0f06c719c54a7ee1e87cb3cb8444dc080635d to make sure that `process.go` `run` manages its own state without the `project_runner.go` involvement and added another test based on your description.

PR #462, opened 2026-04-18
feat: add 'process-compose analyze critical-chain' subcommand

ryantm:
Note this code was generated with AI assistance.

## Summary

Adds a new top-level `analyze` command, with a first subcommand
`analyze critical-chain`, inspired by `systemd-analyze critical-chain`.
It prints a tree of processes from the ones nothing depends on, down
through their dependencies, annotated with startup timings.

For each process two times are shown:

- `@<offset>` — time after the project started that the process became ready
  (or was launched, for processes with no readiness signal).
- `+<duration>` — time the process spent between launch and becoming ready.
  Only shown for processes with a readiness probe / liveness probe /
  `ready_log_line`.

### Example

```
$ process-compose analyze critical-chain

The time when unit became ready is printed after the "@" character.
The time the unit took to become ready is printed after the "+" character.

Project: example/compose
Started: 2026-04-18T09:01:20-07:00
Up time: 4m48.517s

app (not started) [Pending]
├─migrations (not started) [Pending]
│ ├─postgres @4min 30.148s +270.145s
│ └─setup @4ms (not ready)
├─postgres @4min 30.148s +270.145s
├─redis @2min 10.446s +130.435s
└─setup @4ms (not ready)
...
```

You can also restrict the output to a specific process (and its sub-chain):

```
$ process-compose analyze critical-chain app
```

## What changed

- **`src/types/process.go`**: three optional `*time.Time` fields on
  `ProcessState` (`process_start_time`, `process_ready_time`,
  `process_end_time`). All are `omitempty`, so the JSON wire format is
  backward-compatible for existing clients.
- **`src/app/process.go`**: populates the new timestamps at existing
  lifecycle hook points:
  - `run()` — sets `ProcessStartTime`. For processes with no readiness
    signal (no readiness probe and no `ready_log_line`) this is also used
    as `ProcessReadyTime`.
  - `setProcHealth(Ready)` — sets `ProcessReadyTime` when a readiness
    probe succeeds.
  - `handleOutput()` — sets `ProcessReadyTime` when `ready_log_line`
    matches.
  - `onProcessEnd()` — sets `ProcessEndTime`.
- **`src/cmd/analyze.go`** — new parent `analyze` Cobra command.
- **`src/cmd/analyze_critical_chain.go`** — `critical-chain [process...]`
  subcommand. Implemented purely on the client side: composes the
  existing `/project/state`, `/processes`, and `/graph` endpoints, walks
  the dependency graph, and sorts siblings by descending ready-time
  (same order `systemd-analyze` uses). **No new API endpoints.**

## Motivation

When a process-compose project has many processes with deep dependency
chains, it can be hard to see *which* process is the startup bottleneck.
`process list` shows status and age, but doesn't show how long each
process took to reach the state that unblocks its dependents, and
doesn't walk the dependency graph.

`analyze critical-chain` answers the question "what was the slow path?"
at a glance.

## Test plan

- Built locally with `go build -v -o process-compose .` on Go 1.26.
- `go vet ./...` clean.
- Ran against a real project with many processes and a mixture of
  readiness-probe / `ready_log_line` / no-probe processes; confirmed:
  - Top-level processes are those nothing depends on.
  - Siblings sort by descending ready-time (slowest first), matching
    `systemd-analyze`.
  - `@` and `+` values match the expected wall-clock deltas.
  - Processes that never became ready are annotated `(not ready)` /
    `(not started)` and don't crash the output.
  - `[Skipped]`, `[Error]`, `[Restarting]`, `[Launching]`, etc. status
    annotations are surfaced.
  - Passing explicit process names restricts the tree to those
    sub-chains.
  - Tested with a read-only unix-socket client as well as TCP.

## Backward compatibility

- JSON additions are all `omitempty` pointers, so existing API consumers
  see no difference.
- No existing commands or flags were changed.
- No new API endpoints are introduced.

## Open questions

Happy to iterate on any of these:

- Naming: `analyze critical-chain` mirrors `systemd-analyze`. Open to
  `analyze chain`, `analyze timing`, etc. if preferred.
- Color output: currently uses `fatih/color` like other commands
  (`state`, `list`). Can add a `--no-color` flag / respect `NO_COLOR` if
  desired.
- Output format: currently text-only. A `-o json` mode would be a
  straightforward follow-up if there's interest.

### DIFF
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -141,9 +141,20 @@ loop:
 			return 1
 		}
 
-		p.setStartTime(time.Now())
+		now := time.Now()
+		p.setStartTime(now)
 		p.stateMtx.Lock()
 		p.procState.Pid = p.command.Pid()
+		if p.procState.ProcessStartTime == nil {
+			startCopy := now
+			p.procState.ProcessStartTime = &startCopy
+			// Processes without any readiness signal are considered ready as
+			// soon as they start. Probes/log-ready handlers override this later.
+			if p.readyProber == nil && p.procConf.ReadyLogLine == "" {
+				readyCopy := now
+				p.procState.ProcessReadyTime = &readyCopy
+			}
+		}
 		p.metricsProc, err = puproc.NewProcess(int32(p.procState.Pid))
 		if err != nil {
 			log.Err(err).Msgf("Could not find pid %d with name %s", p.procState.Pid, p.getName())
@@ -549,6 +560,13 @@ func (p *Process) onProcessEnd(state string) {
 	p.setState(state)
 	p.updateProcState()
 
+	p.stateMtx.Lock()
+	if p.procState.ProcessEndTime == nil {
+		now := time.Now()
+		p.procState.ProcessEndTime = &now
+	}
+	p.stateMtx.Unlock()
+
 	p.Lock()
 	p.done = true
 	p.Unlock()
@@ -730,6 +748,10 @@ func (p *Process) handleOutput(pipe io.ReadCloser, output string, handler func(m
 		}
 		if p.procConf.ReadyLogLine != "" && p.procState.Health == types.ProcessHealthUnknown && strings.Contains(line, p.procConf.ReadyLogLine) {
 			p.procState.Health = types.ProcessHealthReady
+			if p.procState.ProcessReadyTime == nil {
+				now := time.Now()
+				p.procState.ProcessReadyTime = &now
+			}
 			p.cancelReadyLogFunc(nil)
 		}
 		p.checkElevatedProcOutput(line)
@@ -1063,6 +1085,10 @@ func (p *Process) setProcHealth(health string) {
 	p.stateMtx.Lock()
 	defer p.stateMtx.Unlock()
 	p.procState.Health = health
+	if health == types.ProcessHealthReady && p.procState.ProcessReadyTime == nil {
+		now := time.Now()
+		p.procState.ProcessReadyTime = &now
+	}
 }
 
 // set elevated process password
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -747,11 +747,9 @@ func (p *Process) handleOutput(pipe io.ReadCloser, output string, handler func(m
 			break
 		}
 		if p.procConf.ReadyLogLine != "" && p.procState.Health == types.ProcessHealthUnknown && strings.Contains(line, p.procConf.ReadyLogLine) {
-			p.procState.Health = types.ProcessHealthReady
-			if p.procState.ProcessReadyTime == nil {
-				now := time.Now()
-				p.procState.ProcessReadyTime = &now
-			}
+			// setProcHealth takes stateMtx, updates Health and (for Ready)
+			// records ProcessReadyTime atomically.
+			p.setProcHealth(types.ProcessHealthReady)
 			p.cancelReadyLogFunc(nil)
 		}
 		p.checkElevatedProcOutput(line)
--- a/src/cmd/analyze.go
+++ b/src/cmd/analyze.go
@@ -1,12 +1,12 @@
 package cmd
 
 import (
-	"os"
-
 	"github.com/spf13/cobra"
 )
 
-// analyzeCmd represents the analyze command group.
+// analyzeCmd represents the analyze command group. It has no Run of its own;
+// Cobra auto-prints help when a parent command with no Run is invoked without
+// a subcommand.
 var analyzeCmd = &cobra.Command{
 	Use:   "analyze",
 	Short: "Analyze startup timing and dependency information",
@@ -16,12 +16,6 @@ through the dependency graph.
 
 Available subcommands:
   critical-chain    Print the dependency chains ordered by startup time.`,
-	Run: func(cmd *cobra.Command, args []string) {
-		if len(args) == 0 {
-			_ = cmd.Help()
-			os.Exit(0)
-		}
-	},
 }
 
 func init() {
--- a/src/cmd/analyze.go
+++ b/src/cmd/analyze.go
@@ -1,0 +1,29 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+// analyzeCmd represents the analyze command group.
+var analyzeCmd = &cobra.Command{
+	Use:   "analyze",
+	Short: "Analyze startup timing and dependency information",
+	Long: `Analyze inspects a running process-compose instance to provide insight
+into how long each process took to start up and how those times relate
+through the dependency graph.
+
+Available subcommands:
+  critical-chain    Print the dependency chains ordered by startup time.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			_ = cmd.Help()
+			os.Exit(0)
+		}
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(analyzeCmd)
+}
--- a/src/cmd/analyze_critical_chain.go
+++ b/src/cmd/analyze_critical_chain.go
@@ -1,0 +1,279 @@
+package cmd
+
+import (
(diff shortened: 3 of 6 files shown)


src/cmd/analyze_critical_chain.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// Join with spaces.
+	out := ""
+	for i, p := range parts {
+		if i > 0 {
+			out += " "
+		}
+		out += p
+	}
+	return out
```
F1bonacc1, line 244:
```suggestion
	return strings.Join(parts, " ")
```

src/cmd/analyze.go
```diff
@@ -0,0 +1,29 @@
...
+through the dependency graph.
+
+Available subcommands:
+  critical-chain    Print the dependency chains ordered by startup time.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) == 0 {
+			_ = cmd.Help()
+			os.Exit(0)
+		}
+	},
```
F1bonacc1, line 24:
```suggestion
```
Cobra auto-prints help when a parent command has no Run and no subcommand is provided

src/cmd/analyze_critical_chain.go
```diff
@@ -0,0 +1,279 @@
...
+		return time.Duration(1<<62 - 1)
+	}
+	if s.ProcessReadyTime != nil {
+		return s.ProcessReadyTime.Sub(projectStart)
+	}
+	if s.ProcessStartTime != nil {
+		return s.ProcessStartTime.Sub(projectStart)
+	}
+	return time.Duration(1<<62 - 1)
+}
```
F1bonacc1, line 137:
```suggestion
func readyOffsetForSort(s *types.ProcessState, projectStart time.Time) time.Duration {
    const unreadySortOrder = time.Duration(math.MaxInt64)
	if s == nil {
		return unreadySortOrder
	}
	if s.ProcessReadyTime != nil {
		return s.ProcessReadyTime.Sub(projectStart)
	}
	if s.ProcessStartTime != nil {
		return s.ProcessStartTime.Sub(projectStart)
	}
	return unreadySortOrder
}
```

src/cmd/analyze_critical_chain.go
```diff
@@ -0,0 +1,279 @@
...
+		stateByName[s.Name] = s
+	}
+
+	// Determine the starting set of nodes.
+	var roots []*types.DependencyNode
+	if len(args) > 0 {
+		for _, name := range args {
+			node, ok := graph.AllNodes[name]
+			if !ok {
+				fmt.Fprintf(os.Stderr, "unknown process (or process has no dependencies): %s\n", name)
```
F1bonacc1, line 74: That's true for the graph, but if the user does name an existing isolated process, the CLI could still render a timings line for it (you already do this when `len(args) == 0`). Consider:
```go
node, ok := graph.AllNodes[name]                                                                                                                                                                                                                                                       
  if !ok {                                                                                                                                                                                                                                                                               
      if _, exists := stateByName[name]; exists {
          node = &types.DependencyNode{Name: name}                                                                                                                                                                                                                                       
      } else {                                                                                                                                                                                                                                                                           
          fmt.Fprintf(os.Stderr, "unknown process: %s\n", name)
          os.Exit(1)                                                                                                                                                                                                                                                                     
      }                                                                
  }                                                                                                                                                                                                                                                                                      
  roots = append(roots, node)
```

src/app/process.go
```diff
@@ -730,6 +748,10 @@ func (p *Process) handleOutput(pipe io.ReadCloser, output string, handler func(m
 		}
 		if p.procConf.ReadyLogLine != "" && p.procState.Health == types.ProcessHealthUnknown && strings.Contains(line, p.procConf.ReadyLogLine) {
 			p.procState.Health = types.ProcessHealthReady
+			if p.procState.ProcessReadyTime == nil {
+				now := time.Now()
+				p.procState.ProcessReadyTime = &now
+			}
```
F1bonacc1, line 754: We should protect this with the stateMtx, like you already do in other places, the easiest fix is just to reuse `setProcHealth`.
```suggestion
p.setProcHealth(types.ProcessHealthReady)
```

F1bonacc1 (commented):
Thanks @ryantm,

This can be a super useful feature for all PC users.
I left a few minor comments. 
Please run `make testrace` after you address those, as there is one potential race condition that I flagged.
Do you think you can add some tests to cover the most critical sections?

BTW, I like it that you reused all the existing APIs, especially the graph!

ryantm:
ai generated message:

Thanks for the review @F1bonacc1! Pushed b21614e with the following changes:

- **Race condition in `handleOutput`** — replaced the direct `p.procState.Health = ...` + inline `ProcessReadyTime` mutation with a call to `p.setProcHealth(types.ProcessHealthReady)`, as you suggested. That path now takes `stateMtx` like everywhere else.
- **`analyze` parent `Run`** — removed. Cobra auto-prints help for a parent with no `Run` + no subcommand. (Dropped the `os` import too.)
- **`strings.Join`** — replaced the manual join loop in `formatNodeLine`.
- **`unreadySortOrder` const** — extracted `const unreadySortOrder = time.Duration(math.MaxInt64)` and used it in both branches of `readyOffsetForSort`.
- **Isolated processes as positional args** — if the name isn't in the graph but *is* known to the project state (i.e. it's a valid isolated process), we now synthesize a bare `DependencyNode{Name: name}` and render its timings instead of erroring out.

### Tests

I refactored `renderCriticalChain(w io.Writer, …)` and `sortRootsByReadyTime(...)` out of the run command so they're unit-testable without a server, then added:

- `src/cmd/analyze_critical_chain_test.go`
  - `TestFormatDuration` / `TestFormatOffset` — formatting across ns / us / ms / sub-second / seconds / minute boundaries, plus negative-duration fallback.
  - `TestReadyOffsetForSort` — asserts `ProcessReadyTime` beats `ProcessStartTime`, start-only fallback, and both nil-sentinels return `unreadySortOrder`.
  - `TestFormatNodeLine` — all branches (no state, ready w/ readiness gap, ready w/o gap, launched-but-not-ready, never-started, Error/Skipped/Completed status annotations).
  - `TestSortRootsByReadyTime` — descending ready-time ordering with unready floating to the top and alphabetical tiebreaks.
  - `TestRenderCriticalChainTree` — end-to-end tree walk with a web → app → {db, cache} dependency graph, verifying indentation, branch characters, and per-node timing annotations.
  - `TestRenderCriticalChainSortsSiblingsSlowestFirst` — sibling ordering inside the tree walk.
- `src/app/process_timing_test.go`
  - `TestSetProcHealth_SetsReadyTimeOnce` — `ProcessReadyTime` set on first Ready and **not** overwritten on later Ready calls.
  - `TestSetProcHealth_NotReadyDoesNotSetTime` — NotReady transitions don't touch `ProcessReadyTime`.
  - `TestSetProcHealth_ConcurrentReadyNoRace` — 32 goroutines racing to call `setProcHealth(Ready)`, intended to trip under `go test -race` if the mutex regresses.

### Local verification

```
go build ./...
go vet ./...
go test ./...         # all packages pass
go test -race ./...   # equivalent to `make testrace`, all pass
./bin/golangci-lint run -c .golangci.yaml   # 0 issues
```

Glad you like the graph reuse! Let me know if you'd like anything else tightened up.

ryantm:
ai comment:

Now run literally via make testrace — all packages pass cleanly under the race detector:

src/admitter ok
src/api ok
src/app ok (17.9s — this is where the race fix lives, and it passes)
src/cmd ok (includes the new analyze_critical_chain_test.go)
src/health / loader / mcp / pclog / scheduler / templater / tui / types / updater / util — all ok
No races detected, no test failures. Earlier I had also used go test -race -count=1 ./... which is the same command with cache-busting; the results match.

F1bonacc1 (approved):
Good job @ryantm!
Thanks again!

Issue #460, opened 2026-04-16
Question about MCP


ryantm:
The new MCP feature is cool that it lets you make MCPs out of your own processes. It would be cool if it also had a way to turn on an MCP that lets ai-agents manage process-compose itself and look at logs. I know there is an API and a CLI already, but the structure and plug and play of MCP is nice. Does it already exist? Is this something you will consider?

F1bonacc1:
Hey @ryantm,

All the tooling is already there, and I have been considering it for a while.
Now, when I know that this can be useful, I think I will give it a go.
Stay tuned... :)

F1bonacc1:
Added in v1.110.0

PR #479, opened 2026-05-05
Add BM25 search and dependency graph as tools in the MCP server

jeanlucthumm:
Adds two extra functions to the new MCP server:

- `pc_process_logs_search` which does semantic search across logs. This is helpful for e.g. very spammy logs where tailing would overwhelm the context, or to search across processes.
- `pc_project_dependency_graph` which shows deps between processes. Useful for e.g. diagnosing startups order related issue or simply understanding why a process is waiting to boot.

See #460

### DIFF
--- a/src/mcp/bm25.go
+++ b/src/mcp/bm25.go
@@ -1,0 +1,175 @@
+package mcp
+
+// Hand-rolled Okapi BM25 for search_logs.
+//
+// The Go ecosystem does not have a lightweight, actively maintained BM25
+// package. The closest options are either (a) full-text search engines like
+// blevesearch/bleve, which pull in a large dependency tree we don't need for
+// ranking a few thousand log lines, or (b) small single-maintainer forks with
+// little usage or upkeep. Given BM25 itself is ~100 lines of standard formula,
+// implementing it here keeps go.mod clean and avoids taking on a dependency
+// we'd have to vet and track.
+
+import (
+	"math"
+	"sort"
+	"strings"
+	"unicode"
+)
+
+// Default Okapi BM25 parameters.
+const (
+	DefaultK1 = 1.5
+	DefaultB  = 0.75
+)
+
+// Hit is a single scored document result from a BM25 search.
+type Hit struct {
+	DocID int
+	Score float64
+}
+
+// Corpus holds pre-tokenized documents and the statistics needed to score
+// queries using Okapi BM25.
+type Corpus struct {
+	docs     [][]string       // tokenized docs (one slice per doc)
+	docLens  []int            // doc length per doc
+	avgDL    float64          // average doc length across the corpus
+	docFreq  map[string]int   // term -> number of docs containing term
+	idfCache map[string]float64
+	termFreq []map[string]int // per-doc term frequency
+	k1       float64
+	b        float64
+}
+
+// NewCorpus builds a scoring corpus over the provided tokenized documents.
+// Pass k1 and b as 0 to use the Okapi defaults (1.5 and 0.75 respectively).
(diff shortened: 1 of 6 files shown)


src/mcp/bm25.go
```diff
@@ -0,0 +1,175 @@
...
+	for _, f := range fields {
+		clean := stripNonAlnum(f)
+		if clean != "" {
+			out = append(out, clean)
+		}
+	}
+	return out
+}
+
+func stripNonAlnum(s string) string {
```
F1bonacc1: Are you sure we want to collapse to a single string here?
In case of IPs or paths, this can be a real mess: 192.168.1.1 → 19216811
Have you considered spliting instead? So that `strings.FieldsFunc(s, notAlnum)` would turn `foo.go:42` into `["foo","go","42"]`

src/mcp/control_tools.go
```diff
@@ -296,3 +316,113 @@ func (s *Server) handleProjectIsReady(_ context.Context, _ mcp.CallToolRequest)
 	}
 	return mcp.NewToolResultJSON(result)
 }
+
+// Default and max bounds for pc_process_logs_search.
+const (
+	searchDefaultTopK     = 20
+	searchMaxTopK         = 100
+	searchDefaultLogLimit = 500
+	searchMaxLogLimit     = 5000
```
F1bonacc1, line 325: Do we also want to set a total line limit?
Someone who passes an empty process name (meaning all processes) and a large project with ~50 processes, will result in a corpus with 250K lines tokenized + scored per query...
LLM that calls this on every turn could feel it.
Consider a hard ceiling on total lines (e.g. cap total corpus at 20k lines and warn in the response when truncated).

F1bonacc1 (commented):
Thank you @jeanlucthumm!
Solid PR overall: focused scope, well-commented, good test coverage.
There are a few things that I flagged for your review, just to make sure they weren't missed.

src/mcp/control_tools.go
jeanlucthumm: Did a cap and then a fair-share mechanism so it distributes the cap evenly among all processes.

src/mcp/bm25.go
jeanlucthumm: now splitting by every non-letter / non-digit

F1bonacc1:
That's great @jeanlucthumm,

The PR is ready for merge.
Just one small request - can you please document the new tools here:
https://github.com/F1bonacc1/process-compose/blob/main/www/docs/mcp-server.md#built-in-control-tools

jeanlucthumm:
@F1bonacc1 done!

www/docs/mcp-server.md
```diff
@@ -247,8 +247,10 @@ When enabled, the MCP server starts even if no process has an `mcp:` section, an
 | `pc_process_ports` | Get TCP/UDP ports a process is listening on | `name: string` |
 | `pc_process_logs` | Fetch the most recent log lines (one-shot) | `name: string`, `tail: integer` (default 100), `offset_from_end: integer` (default 0) |
 | `pc_process_logs_truncate` | Truncate the log buffer for a process | `name: string` |
+| `pc_process_logs_search` | Search log buffers using BM25 ranking across one or all processes | `query: string`, `name: string` (optional, all if omitted), `top_k: integer` (default 20, max 100), `log_limit: integer` (default 500, max 5000) |
```
F1bonacc1, line 250: Do we want to mention the 50K total limit and the truncation bool in response?

jeanlucthumm: added note at bottom since it would crowd the table

F1bonacc1:
Superb!
Merging

PR #496, opened 2026-05-25
feat: Automate replica names expansion in DependsOn

pcastellazzi:


pcastellazzi:
I have two questions about this:

- was the Copy intentional here? [https://github.com/F1bonacc1/process-compose/blob/main/src/loader/mutators.go#L134]?

- I went for a fully copy of DependsOnConfig because i am not sure if maps.Clone was enough. I think this is needed if the user makes a change to a single replica from the UI, but i am not 100% sure if that's correct.

```go
    if proc.DependsOn != nil {
        newProc.DependsOn = make(types.DependsOnConfig)
        for k, v := range proc.DependsOn {
            newDep := v
            if v.Extensions != nil {
                newDep.Extensions = maps.Clone(v.Extensions)
            }
            newProc.DependsOn[k] = newDep
        }
    }
```

src/loader/mutators_test.go
```diff
@@ -288,6 +288,40 @@ func Test_cloneReplicas(t *testing.T) {
...
+				Name:      "p0",
+				Replicas:  2,
+				DependsOn: types.DependsOnConfig{"p1": {Condition: types.ProcessConditionStarted}},
+			},
+			"p1": {
+				Name: "p1",
+			},
+			"p2": {
+				Name:      "p2",
+				DependsOn: types.DependsOnConfig{"p1": {Condition: types.ProcessConditionStarted}, "p0-1": {Condition: types.ProcessConditionHealthy}},
```
F1bonacc1: None of the processes here depend on the bare group name `p0 - p2` references the explicit replica `p0-1`, and `p1` isn't replicated.
I ran this under coverage: the expansion loop and the override guard (`mutators.go:137-140`) never execute, so the test passes even if that whole branch is deleted. The `p0-1` assertion below also goes through the plain else path, not the guard it looks like it's exercising (the guard only fires when a process depends on both p0 and p0-1).

F1bonacc1 (commented):
Checked out the branch — builds, `./src/loader` passes under `-race`, and I confirmed the expansion works: `depends_on: p0` on a 2-replica `p0` expands to `p0-0` + `p0-1`, with an explicit `p0-1` override keeping its own condition.

**Backward compatibility looks solid.** Configs that already list explicit replica names (`depends_on: {p0-0, p0-1}`) pass through unchanged — `groupReplicas` is keyed by the original process name, so replica-name entries miss the lookup and fall through the `else` branch untouched. The override guard also handles the mixed case (bare group + explicit replica): the explicit entry wins, the bare name is dropped, no duplication.

Two things worth addressing:

**1. The test doesn't actually cover the new feature.** I ran it under coverage and the `len(replicas) > 1` expansion loop and the override guard (`mutators.go:137-140`) never execute — the test passes even if that whole branch is deleted. The fixture has no process depending on a bare multi-replica group name: `p2` depends on `p1` (single replica) and on `p0-1` (an explicit replica name), so every assertion goes through the `else` branch and the deep-copy. The `p0-1` condition check looks like it's exercising the guard, but it's preserved via the plain `else` path (the guard only fires when a process depends on both `p0` and `p0-1`).

Suggest adding a process that depends on the bare group name plus an explicit replica:
```go
"p3": {
    Name: "p3",
    DependsOn: types.DependsOnConfig{
        "p0":   {Condition: types.ProcessConditionCompleted}, // should expand to p0-0, p0-1
        "p0-1": {Condition: types.ProcessConditionHealthy},   // explicit, should win
    },
},
```
and assert `p3` ends up with `p0-0` (Completed) + `p0-1` (Healthy) and no bare `p0` key. A second case asserting explicit replica names pass through unchanged would lock in backward-compat too.

**2. Worth a docs note.** Since depending on a replicated process used to be a load error and now works (meaning "wait for all replicas" under the given condition), a short mention in the `depends_on` section of `www/docs/launcher.md` would help.

Nice catch on the `Vars` fix, by the way — the old `maps.Copy(newProc.Vars, proc.Vars)` was a no-op (after the shallow struct copy `newProc.Vars` aliased `proc.Vars`, so it copied the map into itself), which meant all replicas shared one `Vars` map. `maps.Clone` fixes that.

F1bonacc1:
Good questions.

>was the Copy intentional here? [https://github.com/F1bonacc1/process-compose/blob/main/src/loader/mutators.go#L134]?

**1.** No, that `maps.Copy` wasn't doing anything useful - after `newProc := *proc`, `newProc.Vars` and `proc.Vars` are the same map, so it copied the map into itself (a no-op), and all replicas ended up sharing one `Vars` map. Your switch to `maps.Clone` is the right fix.

>I went for a fully copy of DependsOnConfig because i am not sure if maps.Clone was enough. I think this is needed if the user makes a change to a single replica from the UI, but i am not 100% sure if that's correct.

**2.** Your instinct is right that `maps.Clone` isn't a full deep copy - it's shallow, so each `ProcessDependency.Extensions` sub-map would stay shared. Your per-entry copy makes those independent. Two caveats though: the rewrite loop rebuilds `DependsOn` from scratch anyway, so the copy is mostly overwritten; and the expansion step (`newDependsOn[replicaName] = depConf`) re-shares one `Extensions` map across all expanded entries of a process - so for true independence you'd clone in that loop too. As for the UI single-replica edit: I traced the runtime paths and nothing mutates `DependsOn[x].Extensions` in place (`UpdateProcess` replaces the whole config), so shared Extensions wouldn't leak across replicas today. Net: the full copy is fine and harmless, just not strictly required for current correctness.

pcastellazzi:
I fixed the tests as you requested, my bad for not checking the coverage my self, sorry about that.
There is a commit with fixes to the clone steps, to include Extensions (ProcessDependency) too for completness.
Let me know what you think.

Regards,
Pablo C.

www/docs/launcher.md
```diff
@@ -42,6 +42,52 @@ processes:
...
+  consumer:
+    command: "/some/binary"
+    replicas: 2
+
+  # Older versions (<= v1.110.0) require the dependencies to be manually expanded
+  producer:
+    command: "/some/other/binary"
+    depends_on:
+      consumer-0: process_healthy
+      consumer-1: process_healthy
```
F1bonacc1:
```suggestion
    depends_on:
      consumer-0:
        condition: process_healthy
      consumer-1:
        condition: process_healthy
```

www/docs/launcher.md
```diff
@@ -42,6 +42,52 @@ processes:
...
+processes:
+  consumer:
+    command: "/some/binary"
+    replicas: 2
+
+  # Newer versions (> v1.110.0) allow the group name to be used directly
+  producer:
+    command: "/some/other/binary"
+    depends_on:
+      consumer: process_healthy
```
F1bonacc1:
```suggestion
    depends_on:
      consumer:
        condition: process_healthy
```

www/docs/launcher.md
```diff
@@ -42,6 +42,52 @@ processes:
...
+  consumer:
+    command: "/some/binary"
+    replicas: 2
+
+  # Also we can set conditions for each instance in the group independently
+  producer:
+    command: "/some/other/binary"
+    depends_on:
+      # set the defaults for the group
+      consumer: process_healthy
```
F1bonacc1:
```suggestion
    depends_on:
      # set the defaults for the group
      consumer:
        condition: process_healthy
```

www/docs/launcher.md
```diff
@@ -42,6 +42,52 @@ processes:
...
+
+      # override a single instance
+      consumer-0: process_started
```
F1bonacc1:
```suggestion
      consumer-0:
        condition: process_started
```

src/loader/mutators.go
```diff
@@ -124,23 +126,67 @@ func cloneReplicas(p *types.Project) {
...
 	// 1. Create a copy of the struct (Shallow copy)
 	newProc := *proc
 
 	// 2. DEEP COPY the Vars Map
-	maps.Copy(newProc.Vars, proc.Vars)
+	newProc.Vars = maps.Clone(proc.Vars)
 
-	// 3. DEEP COPY the Environment Slices
+	// 3. DEEP COPY the DependsOn Map
+	proc.DependsOn = cloneDependencies(proc.DependsOn)
```
F1bonacc1:
```suggestion
	newProc.DependsOn = cloneDependencies(proc.DependsOn)
```

www/docs/launcher.md
```diff
@@ -42,6 +42,52 @@ processes:
     replicas: 2
 ```
 
+When replicas are present (`processes.process_name.replicas > 1`), other processes can depended on the group (`process-name`) or some of its children (`process-name-N` where `N` is a value between 0 and the number of `replicas` minus one).
```
F1bonacc1:
```suggestion
When replicas are present (`processes.process_name.replicas > 1`), other processes can depend on the group (`process-name`) or specific replicas (`process-name-N` where `N` is a value between 0 and the number of `replicas` minus one).
```

src/loader/mutators_test.go
```diff
@@ -288,6 +288,96 @@ func Test_cloneReplicas(t *testing.T) {
...
+	}
+	if _, ok := p.Processes["t2"].DependsOn["p0"]; !ok {
+		t.Errorf("t2 should depend on p0")
+	}
+	if p.Processes["t2"].DependsOn["p0"].Condition != types.ProcessConditionStarted {
+		condition := p.Processes["t2"].DependsOn["p0"].Condition
+		t.Errorf("t2 should depend on p0 and it should be Started, got %+v", condition)
+	}
+	if p.Processes["t2"].DependsOn["p1-0"].Condition != types.ProcessConditionHealthy {
+		condition := p.Processes["t2"].DependsOn["p0"].Condition
```
F1bonacc1:
```suggestion
		condition := p.Processes["t2"].DependsOn["p1-0"].Condition
```

F1bonacc1 (changes requested):

F1bonacc1 (approved):
Thanks for driving this, @pcastellazzi, and for addressing all the comments!

Cheers! 😄

PR #498, opened 2026-05-25
Add ability to define namespaces via environment variable

### LINKED ISSUE #491: Add ability to set the namespaces via environment
## Feature Request

Add ability to use an environment variable like `PC_NAMESPACES` to set the namespaces which will be started.

#### Use Case:

Setting the namespaces via CLI is fine for most cases. Still we have a big application where you normally don't need to run all parts to work locally. This means most of the time I don't need to run all namespaces. I would like to put the necessary ones into `.env` and then have this being the default.

Still a user would need to be able to override this setting via cli arguments, so of you would run `process-compose -n ...` the environment variable should be ignored. I believe this is the same behaviour currently in place with `PC_CONFIG_FILES`.

#### Proposed Change:

Use `PC_NAMESPACES` as an alternative way to define which namespaces to start.

Everybody with a more complex config and the need to start parts of the processes on a regular bases.

##### vars

I tried to use `vars` like this:
```yaml
vars:
  ENABLE_A: true
  ENABLE_B: false

processes:
  a:
    disabled: {{not ENABLE_A}}
    ...

  b:
    disabled: {{not ENABLE_b}}
    ...
```

...but sadly `disabled` does not allow Go templates.

##### overrides

Also I tried to use something like:
```yaml
processes:
  a:
    disabled: true

  b:
    disabled: true
```

and then use an override like this:

```yaml
processes:
  a:
    disabled: false
```

...but this override does not work (may be a bug?).

##### multiple configs

One could split the config up into multiple parts and then use `PC_CONFIG_FILES` instead. But our config looks more like this:

```yaml
_base: &base
  something: we-want-everywhere

processes:
  a:
    <<: *_base
```

...meaning we cannot reuse the code when talking about multiple files.

##### wrap process-compose

This is what we are actually using currently:

```bash
#!/bin/bash
set -euo pipefail

PROCESS_COMPOSE_ARGS=($@)

# We introduce a new environment variable PC_NAMESPACES, which will be used to
# set the namespaces process-compose will start with. You may override
# PC_NAMESPACES in your local .env file or when running process-compose
# to enable other namespaces.
#
# To be sure we only do this if the user has not passed "-n" or "--namespace"
# via the cli arguments. So defining the namespace via cli arguments just
# works as well.
if [[ ! " ${PROCESS_COMPOSE_ARGS[*]} " =~ [[:space:]](-n|--namespace)[[:space:]] ]]; then
  export PC_NAMESPACES="${PC_NAMESPACES:-a,b}"
  namespaces=(${PC_NAMESPACES//,/ })
  for namespace in ${namespaces[@]}; do
    PROCESS_COMPOSE_ARGS+=("-n" "$namespace")
  done
fi

exec /path/to/bin/process-compose "${PROCESS_COMPOSE_ARGS[@]}"
```

...but this solution is error prone and depends on handling different cases in bash - which is pretty ugly and annoying (and we already might break something with this workaround).

ddanier:
Allow to use comma separated list in environment variable `PC_NAMESPACES` as an alternative to `--namespace`/`-n`.

Fixes: https://github.com/F1bonacc1/process-compose/issues/491

### DIFF
--- a/crates/client/src/cli/cmd/flags.rs
+++ b/crates/client/src/cli/cmd/flags.rs
@@ -29,6 +29,7 @@ pub mod env {
     pub const PORT_NUM: &str = "PC_PORT_NUM";
     pub const DISABLE_TUI: &str = "PC_DISABLE_TUI";
     pub const CONFIG_FILES: &str = "PC_CONFIG_FILES";
+    pub const NAMESPACES: &str = "PC_NAMESPACES";
     pub const SHORTCUTS_FILES: &str = "PC_SHORTCUTS_FILES";
     pub const NO_SERVER: &str = "PC_NO_SERVER";
     pub const SOCKET_PATH: &str = "PC_SOCKET_PATH";
--- a/crates/client/src/cli/cmd/up.rs
+++ b/crates/client/src/cli/cmd/up.rs
@@ -59,8 +59,8 @@ pub struct ProcessComposeFlagsUp {
     #[builder(default = false)]
     pub logs_truncate: bool,
 
-    /// Run only specified namespaces (default: all)
-    #[arg(long = "namespace")]
+    /// Run only specified namespaces (default: all, env: PC_NAMESPACES)
+    #[arg(long = "namespace", env = env::NAMESPACES)]
     #[builder(default = Vec::new())]
     pub namespaces: Vec<String>,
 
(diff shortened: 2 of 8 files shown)


ddanier:
@F1bonacc1 I hope this change is well, as I'm not that deep into the `process-compose` code. Still I thought it shouldn't be hard to add the `PC_NAMESPACES` feature I suggested in https://github.com/F1bonacc1/process-compose/issues/491 - so I tried to do just that ;-)

No AI was used in creating this PR.

F1bonacc1 (approved):
That looks great @ddanier!

Special thanks for not using AI for this 😆

PR #510, opened 2026-06-17
feat(tui): copy-on-select, configurable copy key, clearer Select label

### LINKED ISSUE #509: TUI: make copying logs ergonomic & configurable (copy-on-select, configurable copy key, clearer Select label)
#### Use Case:

Copying log output out of the TUI is one of the most common things people do, but the current flow is hard to discover, hard to use, and impossible to customize:

1. Toggle selection mode with `Ctrl+S`.
2. Highlight the text.
3. Press **`Enter`** to actually copy it.

Step 3 is the sticking point. After turning selection on, highlighting text does nothing visible to the clipboard, and almost everyone's instinct is "I selected it, so it's copied" (terminal-style copy-on-select) or `Ctrl+C`/`Cmd+C`. The "now press Enter" step isn't surfaced until you're already in selection mode, and it's easy to miss. The usual outcome is the "I turned Select on but the highlighted text never makes it to my clipboard" confusion, after which people give up on the TUI and `tail -f` the log files instead.

Related papercuts in the same area:

- **The copy/exit keys are the only TUI keys that aren't configurable.** Every other binding is remappable in `shortcuts.yaml`, but the copy key (`Enter`) and exit key (`Esc`) are hardcoded in `createLogSelectionTextArea` (`src/tui/log-operations.go`). A user who wants, say, a Vim-style `y` to yank has no way to set it.
- **The `Select On` / `Select Off` footer label reads as state, not action.** When you see `Select On`, it reads like "selection is currently on", but it's actually the button you press to *turn it on*. (`log_pretty_print` has the same `Pretty On`/`Pretty Off` ambiguity.)
- **The feature is undocumented.** `www/docs/tui.md` documents `shortcuts.yaml` but never mentions log selection, the copy step, or the OSC 52 fallback.

(Acknowledging the obvious question: `Ctrl+C` can't simply be repurposed for copy because it's already the global quit binding — and forwards `SIGINT` to a focused interactive process — so "just use Ctrl+C" isn't a free move. That's part of why copy-on-select + a configurable copy key is the pragmatic fix.)

#### Proposed Change:

1. **Copy-on-select for the mouse** — completing a mouse selection (drag release) copies it to the clipboard automatically, no key press required. This matches how terminals behave and is what most users expect.
2. **Configurable keyboard copy key** — add a `log_copy` action to the existing shortcuts system, default `Enter` (backward compatible), so the copy key can be rebound:
   ```yaml
   # $XDG_CONFIG_HOME/process-compose/shortcuts.yaml
   shortcuts:
     log_copy:
       shortcut: y   # Vim-style yank instead of Enter
   ```
3. **Clearer footer label** — relabel the toggle from `Select On`/`Select Off` to the action it performs (`Enable Select`/`Disable Select`).
4. **Docs** — document the whole copy-to-clipboard flow (selection mode, copy-on-select, the configurable key, and the OSC 52 fallback).

Anyone who copies logs out of the TUI — i.e. nearly every interactive user — both mouse and keyboard users.

#### Alternative Approaches

- Docs/discoverability only (keep `Enter`, just explain it) — addresses discovery but not ergonomics or customization.
- Add a second hardcoded copy key (e.g. `Ctrl+Y`) instead of wiring it into `shortcuts.yaml` — less flexible.
- Make copy-on-select opt-in behind a config flag rather than default-on — happy to gate it if preferred.

A PR implementing all four points is attached.

saada:
## What

Makes copying logs out of the TUI ergonomic and configurable:

1. **Copy-on-select (mouse):** releasing a mouse drag in selection mode copies the highlight to the clipboard automatically — no `Enter` needed.
2. **Configurable keyboard copy key:** new `log_copy` shortcut, default `Enter` (**fully backward compatible**), so it can be rebound — e.g. `y` for a Vim-style yank.
3. **Clearer footer label:** the selection toggle now reads `Enable Select` / `Disable Select` (the action) instead of `Select On` / `Select Off` (which read as current state).
4. **Docs:** new "Copying Logs to the Clipboard" section in `www/docs/tui.md`, including the OSC 52 fallback — previously undocumented.

```yaml
# $XDG_CONFIG_HOME/process-compose/shortcuts.yaml
shortcuts:
  log_copy:
    shortcut: y   # Vim-style yank instead of Enter
```

## Why

The copy key (`Enter`) and exit key (`Esc`) in selection mode were the only TUI bindings not covered by the configurable shortcuts system, and "turn Select on → highlight → nothing's in my clipboard until I press Enter" is a recurring point of confusion. `Ctrl+C` can't simply be adopted for copy because it's already the global quit binding (and forwards `SIGINT` to a focused interactive process), so copy-on-select plus a configurable copy key is the pragmatic fix.

## How

- **`src/tui/actions.go`** — new `ActionLogCopy` (`log_copy`), default `tcell.KeyCR` in `defaultShortcuts`, and a `ShortCutKeys` entry. Relabels `ActionLogSelection`'s `ToggleDescription`. `log_copy` is intentionally left out of the footer/help order arrays, so the footer layout is unchanged.
- **`src/tui/log-operations.go`** — copy logic extracted into `copyLogSelection(collapse bool)` (ignores empty selections so a stray click/press never clobbers the clipboard). The keyboard path matches the configured key via `isLogCopyEvent` (special key *or* rune); the title hint uses `logCopyShortcut`. A `SetMouseCapture` on the text area copies on `MouseLeftUp`. `Esc` still exits.
- **`src/tui/actions_test.go`** — unit tests for the default (`Enter`) and overrides (`y` rune, `Ctrl-Y` key).
- **`www/docs/tui.md`** — new docs section.

## Behavior notes

- Default keyboard behavior is unchanged: `Enter` still copies, `Esc` still exits.
- Copy-on-select fires only on a deliberate mouse drag release with a non-empty selection, in selection mode. If you'd prefer it gated behind a config flag, easy to add.
- The mouse path keeps the highlight visible after copying; the keyboard path collapses it (as before).

## Testing

- `go build ./...` — clean
- `go test ./...` — pass (incl. new tests)
- `gofmt -l` / `go vet ./src/tui/` — clean

Closes #509

### DIFF
--- a/www/docs/tui.md
+++ b/www/docs/tui.md
@@ -47,6 +47,30 @@ processes:
 
 > :bulb: Too long log lines (above 2^16 bytes long) can cause the log collector to hang.
 
+## Copying Logs to the Clipboard
+
+Toggle **selection mode** with `Ctrl+S` (`Enable Select` in the footer). The log pane then becomes selectable and there are two ways to copy:
+
+- **Mouse:** drag to highlight — releasing the mouse copies the selection automatically (copy-on-select), no key press required.
+- **Keyboard:** highlight with `Shift`+arrows and press `Enter` to copy.
+
+Press `Esc` to leave selection mode.
+
+Process Compose copies through your system clipboard when available and falls back to the [OSC 52](https://www.reddit.com/r/vim/comments/k1ydpn/a_guide_on_how_to_copy_text_from_anywhere/) terminal escape sequence otherwise (your terminal must allow OSC 52). A status message confirms which method was used.
+
+The keyboard copy key is configurable via the `log_copy` shortcut — for example, rebind it to `y` for a Vim-style yank:
+
+```yaml
+# $XDG_CONFIG_HOME/process-compose/shortcuts.yaml
+shortcuts:
+  log_select:
+    shortcut: Ctrl-S # toggle selection mode
+  log_copy:
+    shortcut: Enter  # copy the current selection (default)
+```
+
+> :bulb: `Ctrl+C` cannot be used to copy because it is reserved for quitting Process Compose (and for forwarding `SIGINT` to a focused interactive process).
+
 ## Command Palette
 
 Press `:` to open the command palette — a searchable list of available actions. Type to filter, use arrow keys to navigate, and press `Enter` to select.
--- a/src/tui/actions.go
+++ b/src/tui/actions.go
@@ -20,6 +20,7 @@ const (
 	ActionFollowLog        = ActionName("log_follow")
 	ActionWrapLog          = ActionName("log_wrap")
 	ActionLogSelection     = ActionName("log_select")
+	ActionLogCopy          = ActionName("log_copy")
 	ActionProcessStart     = ActionName("process_start")
 	ActionProcessScale     = ActionName("process_scale")
 	ActionProcessInfo      = ActionName("process_info")
@@ -56,6 +57,7 @@ var defaultShortcuts = map[ActionName]tcell.Key{
 	ActionFollowLog:        tcell.KeyF5,
 	ActionWrapLog:          tcell.KeyF6,
 	ActionLogSelection:     tcell.KeyCtrlS,
+	ActionLogCopy:          tcell.KeyCR,
 	ActionProcessScale:     tcell.KeyF2,
 	ActionProcessInfo:      tcell.KeyF3,
 	ActionProcessSignal:    tcell.KeyCtrlX,
@@ -344,9 +346,12 @@ func newShortCuts() *ShortCuts {
 			},
 			ActionLogSelection: {
 				ToggleDescription: map[bool]string{
-					true:  "Select On",
-					false: "Select Off",
+					true:  "Enable Select",
+					false: "Disable Select",
 				},
+			},
+			ActionLogCopy: {
+				Description: "Copy Selection",
 			},
 			ActionLogPrettyPrint: {
 				ToggleDescription: map[bool]string{
--- a/src/tui/actions_test.go
+++ b/src/tui/actions_test.go
@@ -1,0 +1,65 @@
+package tui
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gdamore/tcell/v2"
+)
+
+func TestDefaultLogCopyShortcut(t *testing.T) {
+	sc := newShortCuts()
+	action, found := sc.ShortCutKeys[ActionLogCopy]
+	if !found {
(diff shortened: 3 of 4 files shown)


www/docs/tui.md
```diff
@@ -47,6 +47,30 @@ processes:
 
 > :bulb: Too long log lines (above 2^16 bytes long) can cause the log collector to hang.
 
+## Copying Logs to the Clipboard
+
+Toggle **selection mode** with `Ctrl+S` (`Enable Select` in the footer). The log pane then becomes selectable and there are two ways to copy:
+
+- **Mouse:** drag to highlight — releasing the mouse copies the selection automatically (copy-on-select), no key press required.
+- **Keyboard:** highlight with `Shift`+arrows and press `Enter` to copy.
```
F1bonacc1, line 55: Keyboard selection doesn't work.
Sounds like a documentation fix, not a bug.

www/docs/tui.md
```diff
@@ -47,6 +47,30 @@ processes:
...
+
+Press `Esc` to leave selection mode.
+
+Process Compose copies through your system clipboard when available and falls back to the [OSC 52](https://www.reddit.com/r/vim/comments/k1ydpn/a_guide_on_how_to_copy_text_from_anywhere/) terminal escape sequence otherwise (your terminal must allow OSC 52). A status message confirms which method was used.
```
F1bonacc1, line 59: The OSC 52 fallback is documented in my other repo: https://github.com/F1bonacc1/glippy#osc-52-fallback
I wouldn't add it, especially with a Reddit link.

F1bonacc1 (changes requested):
Hi @saada,

Thank you for the PR!
I accept that log copy should be a configurable shortcut.
And this is exactly what this PR provides, which is great.
I think "copy on selection" to the main clipboard is unexpected and should be behind a configuration or not at all.

There are also a 2 critical issues that should be addressed.
1. **nil dereference**
https://github.com/F1bonacc1/process-compose/blob/1057510943cfa642cf34cb001bb356350e2da87b/src/tui/actions.go#L236
is not wired. Setting a custom shortcut, I reproduced it with `y` panics here:
https://github.com/F1bonacc1/process-compose/blob/1057510943cfa642cf34cb001bb356350e2da87b/src/tui/actions.go#L251
Since `actionFn` is `nil`.

Since it was missed, it's worth adding a test to catch this before fixing.

2. **`ActionLogCopy` is missing from the help screen**
The `ActionLogCopy` action should be added here:
https://github.com/F1bonacc1/process-compose/blob/1057510943cfa642cf34cb001bb356350e2da87b/src/tui/actions.go#L105

Issue #318, opened 2025-04-02
`--log-file` is overwritten by default


9999years:
I have `$PC_LOG_FILE` set to write `process-compose` logs to a file.

This works well, except that `process-compose` truncates the file when opening it by default.

I'll usually run `process-compose --detached`, which logs to the file correctly, but then if I run `process-compose list --output wide`, it will overwrite the log file, which often contains useful data!

E.g. here's the whole log file, missing previous data:

```
$ command cat .pc/logs/process-compose.log
25-04-02 11:51:41.514 INF Process Compose v1.46.0
```

F1bonacc1:
Hey @9999years,

Thanks for letting me know. That sounds like a bug, but I double-checked in the code, and it looks like `list` and `ls` are correctly markerd as client commands here: https://github.com/F1bonacc1/process-compose/blob/f87744d07985e7e941a272666c8ec3213f8b9e6b/src/config/config.go#L46

I also tried reconstructing that with `list,` and the log wasn't truncated. Maybe it is truncated with a different command than `list`?

9999years:
@F1bonacc1 Ah, interesting! I want to keep the logs within the repository, so I have the `PC_LOG_FILE` environment variable set in my shell to point to a path within the repository. Therefore, it always uses the same path regardless of whether or not I'm running a client command.

Maybe we could add a `PC_CLIENT_LOG_FILE` variable? Seems a little clumsy though...

F1bonacc1:
Yes, that's the issue!
If it runs in the same shell, it will affect both the client and the server.
I don't feel comfortable overriding something that was set by the user by adding the `-client` suffix.
On one hand, it creates confusion, so it sounds like an issue; on the other it works as expected.
All the solutions I can think of feel off:
1. Another env variable for the client (thanks for the suggestion).
2. `--add-pid-to-log` flag that will suffix the log file with the process PID (or time).
3. Document the issue
4. If the env variable is used for the client, add a warning to the log that if it runs from the same shell, the server log will be overwritten.

I am, of course, open to more suggestions and will think more about other options.

9999years:
Sounds good. This issue isn't very urgent (the workaround posted above works well), so I'd be fine just documenting it.

soy-chrislo:
Hey, I put up #516 that documents the `PC_LOG_FILE`/client truncation thing for now (docs only), didn't wanted to touch the actual behavior without your ok first.

[...]

PR #516, opened 2026-07-10
docs(logging): clarify PC_LOG_FILE truncation between server and client commands

soy-chrislo:
Small docs addition to `www/docs/logging.md` clarifying that `PC_LOG_FILE` (or `--log-file`) is shared between the server process and client commands, and that client commands will truncate the file on open if it's the same path the server is using.

Related to #318. Keeping this to docs only for now, will follow up on the issue with some thoughts on the actual fix.

Didn't run the full `make ci` since this only touches a markdown file and `build-nix` isn't available in my environment (no Nix installed), so there's nothing Go-related for it to catch here.

### DIFF
--- a/www/docs/logging.md
+++ b/www/docs/logging.md
@@ -84,6 +84,31 @@ log_configuration:
 
 Default log location: `/tmp/process-compose-$USER.log`
 
+### Client Commands and `PC_LOG_FILE`
+
+Client commands (`down`, `attach`, `list`, `process`, `project`, `namespace`) write to
+their own log file so they don't overwrite the log of a running `process-compose`
+server. By default that's `/tmp/process-compose-$USER-client.log`, next to the
+server's `/tmp/process-compose-$USER.log`.
+
+`PC_LOG_FILE` (and `--log-file`) sets the path for every `process-compose`
+invocation, server and client alike. Since each invocation truncates its log file
+on open, exporting `PC_LOG_FILE` in your shell points both at the same file, and
+the next client command starts it over.
+
+To use custom locations, give the server and the clients separate paths:
+
+```shell
+process-compose up --detached --log-file .pc/logs/process-compose.log
+
+export PC_LOG_FILE=.pc/logs/process-compose-client.log
+process-compose list
+```
+
+The explicit `--log-file` on `up` takes precedence over `PC_LOG_FILE`, so the
+server keeps `process-compose.log` while client commands log to
+`process-compose-client.log`.
+
 By default, the internal log file contains ANSI color codes. You can disable color output using the `--log-no-color` flag or the `PC_LOG_NO_COLOR` environment variable.
 
 ```bash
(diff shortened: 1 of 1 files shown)


F1bonacc1 (commented):
Thanks for taking the time to write this up — the behavior you're describing is real, and it's worth documenting.

I'd like to adjust the framing before merging. As written, the section reads more like a bug report than documentation: "wiping out whatever was logged so far", the conditional caution ("Keep the two paths separate if you're doing both in the same shell"), and the link to the open issue all suggest "known defect, unresolved" rather than "here's how this works." Docs elsewhere in the repo don't link issues, and that context fits better in the PR description than in the published page.

The bigger gap is that the section doesn't mention the default. Client commands already log to a separate `-client` file unless `PC_LOG_FILE`/`--log-file` overrides it, so the shared-path case is something the user opts into — not the norm. Leading with the default makes the whole thing read as documented behavior plus a recipe, instead of a caveat.

Suggested rewrite (same length, 20 lines vs. 18):

````markdown
### Client Commands and `PC_LOG_FILE`

Client commands (`down`, `attach`, `list`, `process`, `project`, `namespace`) write to
their own log file so they don't overwrite the log of a running `process-compose`
server. By default that's `/tmp/process-compose-$USER-client.log`, next to the
server's `/tmp/process-compose-$USER.log`.

`PC_LOG_FILE` (and `--log-file`) sets the path for every `process-compose`
invocation, server and client alike. Since each invocation truncates its log file
on open, exporting `PC_LOG_FILE` in your shell points both at the same file, and
the next client command starts it over.

To use custom locations, give the server and the clients separate paths:

```shell
process-compose up --detached --log-file .pc/logs/process-compose.log

export PC_LOG_FILE=.pc/logs/process-compose-client.log
process-compose list
```

The explicit `--log-file` on `up` takes precedence over `PC_LOG_FILE`, so the
server keeps `process-compose.log` while client commands log to
`process-compose-client.log`.
````

Two smaller things:

- The `###` level is right if this is a subsection of **Process Compose Internal Log**, but then it should sit directly under that heading rather than after the `tail -f` tip that closes the section.
- Headings in `logging.md` use Title Case (`## Log Rotation`, `## Logger Configuration`), so `### Client Commands and PC_LOG_FILE` fits better than sentence case.

Happy to merge once the wording is updated. And no concern about `make ci` here — a markdown-only change has nothing for it to catch.

soy-chrislo:
Thanks for the detailed feedback, really appreciate you taking the time to point out exactly what needed to change. I know this is a small, simple docs change, but I wanted to get the framing right rather than just merge it as-is, so I went back and verified the behavior you described against the actual source (`src/config/config.go`, `src/cmd/root.go`) before touching anything.

I agree with all three points: the original framing did read like a bug report, it should've led with the default (`-client.log`) instead of the shared-path edge case, and the issue link didn't belong in the published doc. I applied your rewrite as-is: Title Case heading, no issue link, moved directly under **Process Compose Internal Log** right after the `Default log location:` line.

One thing I ran into while placing it: with the `###` positioned there, the ANSI-color paragraph and the `tail -f` tip end up rendering (and showing in the ToC) as children of `Client Commands and PC_LOG_FILE`, since nothing separates them from it afterward. Checked the rest of `www/docs/`, and this is the only page with unheaded prose trailing after a `###` inside a `##` section; everywhere else is either flat or fully split into subsections. Didn't want to restructure existing content without checking with you first, so left it as one commit for now. Happy to push a small follow-up splitting that trailing content into its own `###` (e.g. `Log Colors` / `Following the Log`) if you'd rather keep the pattern consistent, otherwise fine to leave as-is.

F1bonacc1 (approved):
Great! Thanks for this

Issue #517, opened 2026-07-10
-n/--namespace should prune the dependency graph


HunterLarco:
## Feature Request

Right now scoping process-compose using namespaces preserves the dependency graph. So if you want to bring up only a subset of processes, they will never start because their dependencies have not started. For example:

```yaml
version: "0.5"

processes:
  foo:
    command: "echo 'foo'"
    namespace: "foo"
    depends_on:
      bar:
        condition: process_completed_successfully

  bar:
    command: "echo 'bar'"
    namespace: "bar"
```

observe that

```sh
# starts and executes both processes in order
process-compose --no-server --keep-project

# starts and executes process bar
process-compose --no-server --keep-project --namespace bar

# foo never starts
process-compose --no-server --keep-project --namespace foo
```

#### Use Case:

At our company we use process-compose to orchestrate our local web development. It brings up vendors, backends, and frontends. At times it's useful to start just a subset. However, because we use `depends_on` to correctly bring up the entire stack, starting just a subset (such as just the backend or just the frontend) can hang if its dependencies are not online. While this is "correct" in a sense, the use case we have is sometimes it's genuinely useful to just start part of the stack because we've manually brought its dependencies online elsewhere as part of development. Currently this is not possible in process-compose.

#### Proposed Change:

When namespacing, dependencies on pruned processes should also be pruned.

#### Who Benefits From The Change(s)?

me 😅

#### Alternative Approaches

- `--no-deps` This exists today but is all-or-nothing: it strips every depends_on, including deps within the selected namespace. So if I select a namespace whose processes legitimately depend on each other, those intra-namespace deps get dropped too and startup ordering breaks. It solves "don't hang" but throws out the correct-ordering baby with the bathwater.
- A dedicated flag instead of changing --namespace semantics e.g. --prune-external-deps / --namespace-scoped-deps. This avoids changing existing behavior for anyone relying on cross-namespace deps being honored, at the cost of one more flag to know about. I'd be happy with either this or making it the default for --namespace.

F1bonacc1:
Thanks for this suggestion @HunterLarco 
I think making it the default is a good idea, especially considering the fact that today this configuration wouldn't run, so there are virtually no backward compatibility issues.

The implementation was straightforward, and it will be a part of the next release.

F1bonacc1:
Released in v1.120.0

HunterLarco:
Thanks for the quick resolution!

PR #518, opened 2026-07-11
feat: -n/--namespace prune the dependency graph

HunterLarco:
As per https://github.com/F1bonacc1/process-compose/issues/517 this PR implements two changes:

1. it modifies the existing behavior of `--namespace` so that dangling `depends_on` statements are pruned while retaining dependency relationships within the namespace.
2. it adds `--strict-namespace` for users who wish to error rather than allowing namespaces to prune unsatisfiable dependencies. This is strictly an improvement over the current functionality which hangs forever.

It's been tested both in unit tests and manually using the example config from https://github.com/F1bonacc1/process-compose/issues/517

```yaml
version: "0.5"

processes:
  foo:
    command: "echo 'foo'"
    namespace: "foo"
    depends_on:
      bar:
        condition: process_completed_successfully

  bar:
    command: "echo 'bar'"
    namespace: "bar"
```

for example

```sh
# starts and executes both processes in order
process-compose --no-server --keep-project

# starts and executes process bar
process-compose --no-server --keep-project --namespace bar

# starts and executes process bar
process-compose --no-server --keep-project --namespace bar --strict-namespace

# starts and executes process foo
process-compose --no-server --keep-project --namespace foo

# fails to start with a dependency error on foo --> bar
process-compose --no-server --keep-project --namespace foo --strict-namespace
```


HunterLarco:
Closed in favor of https://github.com/F1bonacc1/process-compose/commit/c8642c1964967da781b1085e1d59b079cae7b951

PR #521, opened 2026-07-17
🐛 Wait for process group extinction during shutdown

### LINKED ISSUE #520: Process group child survives shutdown when leader exits on SIGTERM
### Description

`process-compose down` can return successfully while a same-process-group child remains alive.

This happens when the managed leader handles SIGTERM and exits, but a child in the same process group ignores SIGTERM. The leader's exit cancels `forceKillOnTimeout`, so the configured SIGKILL escalation never runs even though the process group is still occupied.

### Reproduction

Run a managed shell which installs a SIGTERM exit trap, starts a child which ignores SIGTERM, and waits. Configure `shutdown.timeout_seconds: 1`, then call `process-compose down`.

Observed on v1.116.0 and confirmed unchanged on v1.120.0:

- `down` exits zero;
- the leader exits;
- the child remains alive in the original process group.

### Expected behavior

For the default `shutdown.parent_only: false` behavior, shutdown should track the launched process group rather than only the leader. After the graceful timeout it should SIGKILL the retained process group and wait until the group is empty before reporting success.

I have a regression test and implementation ready which retain the PGID at launch and make timeout handling observe group liveness.

poggersdave:
## Why

Shutdown currently tracks the leader process rather than the process group it launched. If that leader exits on SIGTERM while a same-group child ignores the signal, the timeout is cancelled, `down` reports success, and the child remains alive.

## What changed

The command wrapper now retains the process group ID at launch and exposes group-aware liveness. Shutdown waits on the configured target—parent only or the full group—before deciding whether escalation is required. After SIGKILL, it also waits for the group to disappear instead of returning immediately.

A Unix regression starts a leader which exits cleanly on SIGTERM and a same-group child which ignores it, then proves shutdown does not complete with the child alive. Windows keeps its existing process-tree termination behavior.

## How to verify

```bash
go test ./...
go test -race ./...
go vet ./...
```

Closes #520

<sub>The leader left early; shutdown stayed for the rest of the group.</sub>

### DIFF
--- a/src/app/process.go
+++ b/src/app/process.go
@@ -477,21 +477,35 @@ func (p *Process) stopProcess(withNoRestart bool) error {
 }
 
 func (p *Process) forceKillOnTimeout() error {
-	p.mtxStopFn.Lock()
-	p.waitForStoppedCtx, p.waitForStoppedFn = context.WithTimeout(context.Background(), time.Duration(p.procConf.ShutDownParams.ShutDownTimeout)*time.Second)
-	p.mtxStopFn.Unlock()
-	<-p.waitForStoppedCtx.Done()
-	err := p.waitForStoppedCtx.Err()
-	switch {
-	case errors.Is(err, context.Canceled):
+	timeout := time.Duration(p.procConf.ShutDownParams.ShutDownTimeout) * time.Second
+	if p.waitForCommandExit(timeout, p.procConf.ShutDownParams.ParentOnly) {
 		return nil
-	case errors.Is(err, context.DeadlineExceeded):
-		log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
-		return p.command.Stop(int(syscall.SIGKILL), false)
-	default:
-		log.Error().Err(err).Msgf("terminating %s with timeout %d failed", p.getName(), p.procConf.ShutDownParams.ShutDownTimeout)
+	}
+
+	log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
+	if err := p.command.Stop(int(syscall.SIGKILL), false); err != nil {
 		return err
 	}
+	if p.waitForCommandExit(timeout, false) {
+		return nil
+	}
+	return fmt.Errorf("process group for %s is still running after SIGKILL", p.getName())
+}
+
+func (p *Process) waitForCommandExit(timeout time.Duration, parentOnly bool) bool {
+	deadline := time.NewTimer(timeout)
+	defer deadline.Stop()
+	poll := time.NewTicker(20 * time.Millisecond)
+	defer poll.Stop()
+
+	for p.command.IsAlive(parentOnly) {
+		select {
+		case <-deadline.C:
+			return !p.command.IsAlive(parentOnly)
+		case <-poll.C:
+		}
+	}
+	return true
 }
 
 func (p *Process) doConfiguredStop(params types.ShutDownParams) error {
--- a/src/app/system_test.go
+++ b/src/app/system_test.go
@@ -1216,6 +1216,65 @@ func TestSystem_TestProcShutDownWithConfiguredTimeOut(t *testing.T) {
 
 }
 
+func TestSystem_ShutdownWaitsForProcessGroupExtinction(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("test requires Unix process-group semantics")
+	}
+
+	const processName = "leader_exits_first"
+	childPIDFile := filepath.Join(t.TempDir(), "child.pid")
+	shell := command.DefaultShellConfig()
+	project := &types.Project{
+		Processes: map[string]types.ProcessConfig{
+			processName: {
+				Name:        processName,
+				ReplicaName: processName,
+				Executable:  shell.ShellCommand,
+				Args: []string{
+					shell.ShellArgument,
+					fmt.Sprintf("trap 'exit 0' TERM; bash -c \"trap '' TERM; while true; do sleep 1; done\" & echo $! > %q; wait", childPIDFile),
+				},
+				RestartPolicy: types.RestartPolicyConfig{Restart: types.RestartPolicyNo},
+				ShutDownParams: types.ShutDownParams{
+					ShutDownTimeout: 1,
+					Signal:          int(syscall.SIGTERM),
+				},
+			},
+		},
+		ShellConfig: shell,
+	}
+
+	runner, err := NewProjectRunner(&ProjectOpts{project: project})
+	if err != nil {
+		t.Fatal(err)
+	}
+	go func() { _ = runner.Run() }()
+
+	var childPID int
+	deadline := time.Now().Add(5 * time.Second)
+	for time.Now().Before(deadline) {
+		raw, readErr := os.ReadFile(childPIDFile)
+		if readErr == nil {
+			parsed, _ := fmt.Sscanf(string(raw), "%d", &childPID)
+			if parsed == 1 {
+				break
+			}
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+	if childPID <= 1 {
+		t.Fatal("child process never started")
+	}
+	t.Cleanup(func() { _ = syscall.Kill(childPID, syscall.SIGKILL) })
+
+	if err := runner.StopProcess(processName); err != nil {
+		t.Fatal(err)
+	}
+	if err := syscall.Kill(childPID, 0); !errors.Is(err, syscall.ESRCH) {
+		t.Fatalf("child process %d survived shutdown", childPID)
+	}
+}
+
 func TestSystem_TestRestartingProcessShutDown(t *testing.T) {
 	proc1 := "proc1"
 	shell := command.DefaultShellConfig()
--- a/src/command/command_wrapper.go
+++ b/src/command/command_wrapper.go
@@ -4,10 +4,12 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"sync/atomic"
 )
 
 type CmdWrapper struct {
-	cmd *exec.Cmd
+	cmd            *exec.Cmd
+	processGroupID atomic.Int64
 }
 
 func (c *CmdWrapper) GetPty() *os.File {
@@ -15,7 +17,11 @@ func (c *CmdWrapper) GetPty() *os.File {
 }
 
 func (c *CmdWrapper) Start() error {
-	return c.cmd.Start()
+	if err := c.cmd.Start(); err != nil {
+		return err
+	}
+	c.captureProcessGroup()
+	return nil
 }
 
 func (c *CmdWrapper) Run() error {
--- a/src/command/stopper_unix.go
+++ b/src/command/stopper_unix.go
@@ -3,8 +3,10 @@
 package command
 
 import (
-	"github.com/rs/zerolog/log"
+	"errors"
 	"syscall"
+
+	"github.com/rs/zerolog/log"
 )
 
 const (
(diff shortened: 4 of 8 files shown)


poggersdave:
Live macOS arm64 smoke on commit 24a3652: in a disposable git worktree, the supervised leader exited on SIGTERM while a child in the same process group ignored SIGTERM. `process-compose down` returned success only after both exact fixture PIDs were extinct; the Unix socket and config were then removed. The disposable worktree was deleted afterward, with the pre-existing worktree paths and HEADs unchanged.

src/command/stopper_unix.go
```diff
@@ -31,12 +33,40 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
...
+
+func (c *CmdWrapper) captureProcessGroup() {
 	pgid, err := syscall.Getpgid(c.Pid())
 	if err == nil {
-		return syscall.Kill(-pgid, syscall.Signal(sig))
+		c.processGroupID.Store(int64(pgid))
 	}
-
-	return err
 }
```
F1bonacc1, line 70: **Issue (must fix): this captures process-compose's own pgid for `AttachIo()` processes, which makes the SIGKILL escalation kill us.**

`src/app/process.go:263` calls `SetCmdArgs()` the only place `Setpgid: true` is set, solely in the `else` branch:

```go
if p.isMain || (p.procConf.IsElevated && !p.isTuiEnabled) {
    p.command.AttachIo()
} else {
    p.command.SetCmdArgs()
    ...
}
```

So `up --main <name>` and elevated processes in headless/daemon mode inherit *our* process group, and `Getpgid(c.Pid())` returns our own pgid. I confirmed this on this branch:

```
process-compose(test) pid=28256 pgid=27239
captured child pgid=27239          <- our own pgid
IsAlive(parentOnly=true)  = false  <- leader exited and was reaped
IsAlive(parentOnly=false) = true   <- group "alive": that's us
```

Then `waitForCommandExit(timeout, false)` never succeeds → the full `timeout_seconds` is burned → `Stop(SIGKILL, false)` → `kill(-ownPgid, SIGKILL)` kills process-compose and its whole group, uncatchably. Before this PR the escalation only ran when the leader failed to exit, so it was practically unreachable; now it fires on every shutdown of such a process.

Guarding the capture fixes it, and keeps every group operation from ever targeting us:

```suggestion
func (c *CmdWrapper) captureProcessGroup() {
	pgid, err := syscall.Getpgid(c.Pid())
	if err != nil {
		return
	}
	// Processes started via AttachIo() (isMain / elevated) skip SetCmdArgs, so they
	// inherit our own process group. Group-wide signals and liveness checks would
	// then target process-compose itself.
	if pgid == syscall.Getpgrp() {
		return
	}
	c.processGroupID.Store(int64(pgid))
}
```

With this plus the `IsAlive` fallback below, `go test -race ./src/app/... ./src/command/...` passes and a probe confirms the `AttachIo` case reports the exited process as dead while the normal `Setpgid` case still goes alive → dead.

src/command/stopper_unix.go
```diff
@@ -31,12 +33,40 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
...
+	}
+	if parentOnly {
+		err := c.cmd.Process.Signal(syscall.Signal(0))
+		return err == nil || errors.Is(err, syscall.EPERM)
+	}
+
+	pgid := int(c.processGroupID.Load())
+	if pgid <= 0 {
+		return false
+	}
```
F1bonacc1, line 60: **Issue (must fix): an unknown group is reported as extinct, which silently skips the SIGKILL escalation.**

When `pgid <= 0`, `forceKillOnTimeout`'s first `waitForCommandExit` returns `true` immediately and the function returns `nil` without ever escalating, a regression from the previous behaviour, where the deadline always escalated. This is reachable whenever `Getpgid` fails in `captureProcessGroup`, unconditionally for commands started via `Run()`/`Output()`/`CombinedOutput()` (which never call it), and for every `AttachIo()` process once the guard above is added.

Falling back to the leader is both safer and correct:

```suggestion
	pgid := int(c.processGroupID.Load())
	if pgid <= 0 {
		// Unknown group: fall back to the leader rather than claiming extinction.
		return c.IsAlive(true)
	}
```

src/command/stopper_unix.go
```diff
@@ -31,12 +33,40 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
...
+	pgid := int(c.processGroupID.Load())
+	if pgid <= 0 {
+		var err error
+		pgid, err = syscall.Getpgid(c.Pid())
+		if err != nil {
+			return err
+		}
+		c.processGroupID.CompareAndSwap(0, int64(pgid))
+	}
+	return syscall.Kill(-pgid, syscall.Signal(sig))
```
F1bonacc1, line 45: Same exposure as `captureProcessGroup`: this fallback will happily resolve and signal process-compose's own group for `AttachIo()` processes. Once the `pgid == syscall.Getpgrp()` guard is in place it's worth applying it here too (or routing both through a single `resolveProcessGroup()` helper) so there's exactly one place that decides whether group-wide signalling is safe.

Separately: caching the pgid means `Stop` no longer verifies the leader still exists before signalling, where the old `Getpgid` call would fail with `ESRCH` and send nothing. The exposure is largely the same either way (a recycled PID would have resolved to a wrong group before, too), so this is a note rather than a request.

src/command/stopper_windows.go
```diff
@@ -35,6 +35,14 @@ func (c *CmdWrapper) Stop(sig int, parentOnly bool) error {
 	return nil
 }
 
+func (c *CmdWrapper) IsAlive(_ bool) bool {
+	return c.cmd != nil && c.cmd.ProcessState == nil
+}
```
F1bonacc1, line 40: **Issue (must fix): data race on `cmd.ProcessState`.**

`ProcessState` is written by `exec.Cmd.Wait()`, which runs in the process's `run()` goroutine (`src/app/process.go:178`), while `IsAlive` is called from the shutdown goroutine via `waitForCommandExit`. That's an unsynchronized read/write of the same field, and the `//go:build windows` tag means Linux CI's `-race` will never flag it.

Tracking liveness explicitly avoids touching `Wait`-owned state, e.g. an `atomic.Bool` set in `Start()` and cleared once `Wait()` returns, mirroring how `MockCommand` does it in this PR.

src/app/process.go
```diff
@@ -477,21 +477,35 @@ func (p *Process) stopProcess(withNoRestart bool) error {
...
 		return nil
-	case errors.Is(err, context.DeadlineExceeded):
-		log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
-		return p.command.Stop(int(syscall.SIGKILL), false)
-	default:
-		log.Error().Err(err).Msgf("terminating %s with timeout %d failed", p.getName(), p.procConf.ShutDownParams.ShutDownTimeout)
+	}
+
+	log.Debug().Msgf("process failed to shut down within %d seconds, sending %d", p.procConf.ShutDownParams.ShutDownTimeout, syscall.SIGKILL)
+	if err := p.command.Stop(int(syscall.SIGKILL), false); err != nil {
```
F1bonacc1, line 486: The hardcoded `false` here means a process configured with `shutdown.parent_only: yes` still gets its children SIGKILLed, the children the user explicitly asked to spare. That predates this PR, but since this function now reasons explicitly about signal targets (the first wait correctly honours `ParentOnly`), this is the natural place to make the escalation consistent with the configuration.

src/app/process.go
```diff
@@ -477,21 +477,35 @@ func (p *Process) stopProcess(withNoRestart bool) error {
...
 		return err
 	}
+	if p.waitForCommandExit(timeout, false) {
```
F1bonacc1, line 489: **Suggestion: worst-case shutdown time doubles.**

Reusing the full `timeout` for the post-SIGKILL wait makes a stuck process take up to `2 × shutdown.timeout_seconds` (40s with two defaults), which users will notice on `down`. SIGKILL is synchronous for anything not in uninterruptible sleep, so a short grace is enough:

```suggestion
	if p.waitForCommandExit(min(timeout, 2*time.Second), false) {
```

(Compiles and the suite stays green with this applied; promote the literal to a named const if you prefer.)

src/app/process.go
```diff
@@ -477,21 +477,35 @@ func (p *Process) stopProcess(withNoRestart bool) error {
...
+		return nil
+	}
+	return fmt.Errorf("process group for %s is still running after SIGKILL", p.getName())
```
F1bonacc1, line 492: **This changes the API contract.**

`forceKillOnTimeout` now returns an error where it previously returned `nil`, and that propagates through `runner.StopProcess` to the REST `/process/stop/{name}` handler. A group member in `D` state or just the zombie window described below will make `down`/`stop` report failure for a process that is effectively gone.

A `log.Warn` may serve users better here. If the hard error is deliberate, it's worth a release-note line, since scripts checking the exit status of `down` will start seeing new failures.

src/app/process.go
```diff
@@ -477,21 +477,35 @@ func (p *Process) stopProcess(withNoRestart bool) error {
...
+	defer poll.Stop()
+
+	for p.command.IsAlive(parentOnly) {
+		select {
+		case <-deadline.C:
+			return !p.command.IsAlive(parentOnly)
+		case <-poll.C:
+		}
+	}
+	return true
```
F1bonacc1, line 508: **Suggestion: a zombie leader counts as a live group, so this can escalate against an already-dead process.**

Both `kill(-pgid, 0)` and `Signal(0)` succeed for an exited-but-unreaped process. Measured on this branch:

```
pre-reap : IsAlive(true)=true  IsAlive(false)=true
post-reap: IsAlive(true)=false IsAlive(false)=false
```

Reaping happens in `p.command.Wait()`, which sits behind `p.waitForStdOutErr()` (`process.go:177`) and that wait is *unbounded* (plain `context.WithCancel`) when the process was still `Running` on entry, escaping only via `procRunCtx.Done()`, which `stopProcess` cancels only when `withNoRestart` is true. On the restart path the leader can therefore stay a zombie past the deadline, producing an unnecessary group SIGKILL and possibly the new `"still running after SIGKILL"` error for a process that already exited.

The codebase already has an authoritative leader-exit signal: `waitForStoppedFn`, still wired up in `onProcessEnd` (`process.go:611`) and still used by `doSendKeysStop`. Combining the two (context signal for the leader, polling only for the remaining group members) would be both exact and cheaper than a 20ms poll. Not blocking, but it removes a whole class of timing dependence.

src/app/system_test.go
```diff
@@ -1216,6 +1216,65 @@ func TestSystem_TestProcShutDownWithConfiguredTimeOut(t *testing.T) {
...
+	shell := command.DefaultShellConfig()
+	project := &types.Project{
+		Processes: map[string]types.ProcessConfig{
+			processName: {
+				Name:        processName,
+				ReplicaName: processName,
+				Executable:  shell.ShellCommand,
+				Args: []string{
+					shell.ShellArgument,
+					fmt.Sprintf("trap 'exit 0' TERM; bash -c \"trap '' TERM; while true; do sleep 1; done\" & echo $! > %q; wait", childPIDFile),
```
F1bonacc1, line 1235: **Suggestion: the child can lose this race, making the test pass against unfixed code.**

`echo $! > pidfile` runs in the *parent* immediately after backgrounding, so the PID file can become readable before the inner shell has installed `trap '' TERM`. If SIGTERM lands in that window the child dies from SIGTERM and the assertion holds even without the fix, the regression test quietly stops being one. It reproduces reliably today, but CI under load is exactly where that window widens.

Having the child publish its own PID *after* installing the trap closes it, and using `shell.ShellCommand` instead of a hardcoded `bash` keeps the inner shell consistent with `COMPOSE_SHELL`:

```suggestion
					fmt.Sprintf("trap 'exit 0' TERM; %s -c 'trap \"\" TERM; echo $$ > %q; while true; do sleep 1; done' & wait", shell.ShellCommand, childPIDFile),
```

I verified this rewrite: passes 3/3 on this branch and fails 3/3 on `main`, under both the default `bash` and `COMPOSE_SHELL=sh`.

src/app/system_test.go
```diff
@@ -1216,6 +1216,65 @@ func TestSystem_TestProcShutDownWithConfiguredTimeOut(t *testing.T) {
...
+	}
+	t.Cleanup(func() { _ = syscall.Kill(childPID, syscall.SIGKILL) })
+
+	if err := runner.StopProcess(processName); err != nil {
+		t.Fatal(err)
+	}
+	if err := syscall.Kill(childPID, 0); !errors.Is(err, syscall.ESRCH) {
+		t.Fatalf("child process %d survived shutdown", childPID)
+	}
+}
```
F1bonacc1, line 1276: **Suggestion: the other new branches are untested.**

Nothing covers `shutdown.parent_only: yes`, the new `"still running after SIGKILL"` error path, or `IsAlive` itself `src/command` has no test files at all. A small table test over `IsAlive` across {with `Setpgid`, without `Setpgid`, after reap, zombie} is cheap and would have caught both must-fix items above; that's essentially the probe I used to confirm them.

src/command/command_wrapper.go
```diff
@@ -4,18 +4,24 @@ import (
...
 func (c *CmdWrapper) Start() error {
-	return c.cmd.Start()
+	if err := c.cmd.Start(); err != nil {
+		return err
+	}
+	c.captureProcessGroup()
+	return nil
 }
 
 func (c *CmdWrapper) Run() error {
```
F1bonacc1, line 27: `Start()` captures the process group but `Run()` doesn't, so anything launched through `Run()`/`Output()`/`CombinedOutput()` has `processGroupID == 0` forever. Nothing calls `IsAlive` on those paths today (`doConfiguredStop`, `health/exec_checker.go`, `project_runner.go:1523`), so this is latent, but capturing in both keeps the invariant "started ⇒ group known" simple for the next caller.

F1bonacc1 (changes requested):

F1bonacc1:
Thank you for this PR @poggersdave!
Tracking the launched process group instead of only the leader is the right model, and the regression test genuinely catches the bug.

PR #524, opened 2026-07-24
 feat(completion): dynamic process-name completion for run/up and process start/stop/restart

ilyagr:
Add shell completion to process-name-taking commands such as
`process-compose run <TAB>`. Uses cobra's `ValidArgsFunction`
functionality. Should work in every supported shell.

`run` and `up` load names from the config file, which also allows them
to pass the first line of the process descriptions (if defined in the
config) to `cobra` together with the names.

The `process start/stop/restart` commands query the running server.
The server doesn't currently return the process descriptions (this could
be added relatively easily), so the completions will come without
descriptions. 

On any error (missing config or unreachable server) the helpers return
ShellCompDirectiveNoFileComp with no candidates. The config path sets
IsInternalLoader so a malformed or half-edited config returns an error
instead of aborting completion via log.Fatal.

-----

## About the third commit, `honor --unix-socket ...`

That is a preventive fix to a somewhat obscure but really confusing problem
discovered by AI. See the description and the code for details. It comes with
a somewhat heavy-weight AI-generated test to prove that the bug actually exists.

I'm happy to remove that commit if you prefer a different solution, or to
keep the fix and remove the heavy-weight (but IMO easy-to-follow) test.

One potentially better solution IMO would be to remove the `-U` flag entirely, I'm not
sure it's useful when `--unix-socket` implies `-U` but a bare `-U` creates
a socket at a hard-to-guess path.

### DIFF
--- a/src/cmd/completion_process.go
+++ b/src/cmd/completion_process.go
@@ -1,0 +1,72 @@
+package cmd
+
+import (
+	"strings"
+
+	"github.com/f1bonacc1/process-compose/src/loader"
+	"github.com/spf13/cobra"
+)
+
+// completionEntry formats a process name and its optional description as a
+// cobra completion candidate.
+func completionEntry(name, description string) string {
+	if desc, _, _ := strings.Cut(description, "\n"); desc != "" {
+		// cobra expects "name\tdescription". We replace tabs with spaces in the
+		// description. cobra doesn't currently care about tabs beyond the
+		// first, but spaces are shorter and zsh converts such tabs into `:`s.
+		return name + "\t" + strings.ReplaceAll(desc, "\t", " ")
+	}
+	return name
+}
+
+// completeProcessNamesFromConfig returns a cobra completion function that
+// completes process names by loading the config file(s), without a running
+// server. Used by commands that start their own server (run, up).
+//
+// For processes that have configured descriptions, the names come concatenated
+// with the first line of the descriptions in the cobra-appropriate
+// "name\tdescription" format.
+//
+// When single is true (e.g. `run`, which takes exactly one PROCESS), it stops
+// offering candidates once a positional arg is already present.
+func completeProcessNamesFromConfig(single bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+	return func(_ *cobra.Command, args []string, _ string) ([]string, cobra.ShellCompDirective) {
+		if single && len(args) != 0 {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		// Avoid aborting or printing errors for e.g. unparseable YAML config
+		opts.IsInternalLoader = true
+		project, err := loader.Load(opts)
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		names, err := project.GetLexicographicProcessNames()
+		if err != nil {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		comps := make([]string, len(names))
+		for i, name := range names {
+			comps[i] = completionEntry(name, project.Processes[name].Description)
+		}
+		return comps, cobra.ShellCompDirectiveNoFileComp
+	}
+}
+
+// completeProcessNamesFromServer returns a cobra completion function that
+// completes process names from a running process-compose server. Used by the
+// `process` subcommands (start/stop/restart), which act on a live project; this
+// also surfaces dynamically scaled replica names.
+func completeProcessNamesFromServer(single bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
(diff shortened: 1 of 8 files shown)


src/cmd/completion_process_test.go
```diff
@@ -0,0 +1,187 @@
...
+// `completeProcessNamesFromServer` to work correctly. Without it, completion
+// would still work correctly for `process-compose -U --unix-socket blah process
+// start <TAB>` or `PC_SOCKET_PATH=blah process-compose ...`, but it would
+// surprisingly not work for just `process-compose --unix-socket blah ...`.
+//
+// While few people are likely to use process name completion with
+// `--unix-socket`, they would be really confused if we didn't implement such a
+// workaround, since they'd be used to `--unix-socket` implying `-U` in all
+// other situations.
+func TestProcessStopCompletionOverUnixSocketFlag(t *testing.T) {
```
F1bonacc1, line 129: This test will fail on the `windows-latest` CI leg: the `--unix-socket` flag is only registered when `runtime.GOOS != "windows"` (root.go), and cobra's `getCompletions` errors out on unknown flags during `__complete` (completions.go:373 in cobra v1.10.2), so the command yields no candidates and the `strings.Contains` assertions fail.

Please guard at the top:

```go
if runtime.GOOS == "windows" {
	t.Skip("--unix-socket flag is not registered on Windows")
}
```

(FWIW I verified the test is meaningful on Linux — with the workaround in `completeProcessNamesFromServer` removed, it fails with output `":4\n"`, so it does guard the fix.)

src/cmd/completion_process.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// For processes that have configured descriptions, the names come concatenated
+// with the first line of the descriptions in the cobra-appropriate
+// "name\tdescription" format.
+//
+// When single is true (e.g. `run`, which takes exactly one PROCESS), it stops
+// offering candidates once a positional arg is already present.
+func completeProcessNamesFromConfig(single bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+	return func(_ *cobra.Command, args []string, _ string) ([]string, cobra.ShellCompDirective) {
+		if single && len(args) != 0 {
```
F1bonacc1, line 34: Minor UX regression for `run`: once the process name is present, this returns `NoFileComp`, which also suppresses file completion for the passthrough args after `--` (`process-compose run web -- ./scr<TAB>` no longer completes files; previously the shell fell back to file completion).

Since flags are already parsed by the time `ValidArgsFunction` runs, this branch could return `cobra.ShellCompDirectiveDefault` when `cmd.Flags().ArgsLenAtDash() >= 0` (needs the `cmd` param that's currently discarded) and keep `NoFileComp` otherwise. Fine as a follow-up too.

src/cmd/completion_process.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+// completeProcessNamesFromServer returns a cobra completion function that
+// completes process names from a running process-compose server. Used by the
+// `process` subcommands (start/stop/restart), which act on a live project; this
+// also surfaces dynamically scaled replica names.
+func completeProcessNamesFromServer(single bool) func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+	return func(cmd *cobra.Command, args []string, _ string) ([]string, cobra.ShellCompDirective) {
+		if single && len(args) != 0 {
```
F1bonacc1, line 61: Tiny nit, applies to both helpers when `single` is false: already-typed names are offered again (`process stop web <TAB>` suggests `web` a second time). Filtering out names already present in `args` would be a one-liner. Not blocking.

src/cmd/completion_process.go
```diff
@@ -0,0 +1,77 @@
...
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+		// Normally done in PersistentPreRun so that --unix-socket implies -U; but
+		// during completion PreRun runs before flag parsing, so we must do it here.
+		if isUnixSocketMode(cmd) {
+			*pcFlags.IsUnixSocket = true
+		}
+		// Names only, no descriptions. TODO: consider adding process
+		// descriptions to the server's `/processes` response.
+		names, err := getClient().GetLexicographicProcessNames()
```
F1bonacc1, line 71: Heads-up rather than a request: `getClient()` builds an `http.Client` with no timeout, so completion against a configured-but-unresponsive server (e.g. a firewalled address that drops packets) will hang the user's shell until the kernel TCP timeout. This is pre-existing for all client commands, but completion is more latency-sensitive. A short-timeout client for the completion path would be a nice follow-up; doesn't need to happen in this PR.

F1bonacc1 (changes requested):
Thanks for this solid and well tested PR!
There are a few minor nits and one change request.

